Issue #4, opened 2019-07-29
Allow to customize OpenAPI object programmatically


brunomendola:
The only way to customize the OpenAPI header values is by declaring a bean having the OpenAPIDefinition annotation. But this way the values can't be built programmatically, e.g. I want to build the security schemas by looking at some properties configured in the external configuration file of my Spring application.

PR #5, opened 2019-07-29
Allow to customize OpenAPI object programmatically

brunomendola:
This PR allows to customize the OpenAPI object programmatically, by declaring a bean of type OpenAPI.

Additionally, I started with unit tests :)

Ref. #4


springdoc:
Its also possible to add the API defintion using: @OpenAPIDefinition
Allowing to customize OpenAPI object programmatically, is also a good feature.

PR #56, opened 2019-08-22
Fix Null Safety in  ParameterBuilder.buildParameterFromDoc

antechrestos:
In commit [14ce3c6](https://github.com/springdoc/springdoc-openapi/commit/14ce3c6c7d20cf01d06c2baf2215bd807cb905e9), `schema` may be `null` if `AnnotationsUtils.getSchemaFromAnnotation` returns an empty `Optional`. This change provides a way taking preference in the provided documentation and calculates the schema as a fallback.


antechrestos:
Hi

Let me explain. My fear was that now with the late change schema may be null while it could not be previously. The change brought by the current commit prevent this behavior

antechrestos:
@springdoc I updated the title of PR and the commit message to better reflect the change I brought.  The `null`  result may never occur (I did not read all swagger code) but  this change enforce it to never occur.
Tell me if you are ok, if not I will close the PR.
Thank you for the late change anyway. :smiley:

antechrestos:
Close cause no interaction

PR #101, opened 2019-09-27
Support for x-forward-prefix when using ForwardedHeaderFilter.

nokesc:
I have multiple context roots for the application.  The default context root is **/myapp**, and the other path is **/dev-myapp** using **x-forward-prefix** .  Spring handles this correctly when using **ForwardHeaderFilter**.  GeneralInfoBuilder caches the context root in Server.url when the first request is made to the application.

The OpenAPI object is a singleton, so is impossible to correctly send back the Server.url for individual clients.   I solved my multiple context root problem by creating a new OpenAPI object for each request using a shallow copy and always getting the baseurl value in context of the current request.

Given this implementation, it probably makes more sense to remove lines 70-73 of GeneralInfoBuilder:
```
		// default server value
		if (CollectionUtils.isEmpty(openAPI.getServers())) {
			Server server = new Server().url(serverBaseUrl).description(DEFAULT_SERVER_DESCRIPTION);
			openAPI.addServersItem(server);
		}
```

More generally, I recommend that the OpenAPI template object be calculated during startup only.  Currently the processing to configure OpenAPI is run during each request.

### DIFF
--- a/springdoc-openapi-core/src/main/java/org/springdoc/api/OpenAPICloner.java
+++ b/springdoc-openapi-core/src/main/java/org/springdoc/api/OpenAPICloner.java
@@ -1,0 +1,37 @@
+package org.springdoc.api;
+
+import static org.springdoc.core.Constants.DEFAULT_SERVER_DESCRIPTION;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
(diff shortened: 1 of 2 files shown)


springdoc:
Hi, 

 Thank you for your feedback and your pull request. We will schedule it for a future release.
 Great that you have managed to handle it.

PR #110, opened 2019-10-09
Never ignore @PathVariable parameters

neiser:
Given a controller like
```
@RestController
public class HelloController {
    @GetMapping(path = "/documents/{locale}")
    public ResponseEntity<String> getDocumentsWithLocale(@Parameter(schema = @Schema(type = "string")) @PathVariable("locale") Locale locale) {
        return null;
    }
}
```
where a parameter with type `Locale` is used as a `@PathVariable`, the parameter still ignored although it's required to have a valid spec generated.

This PR fixes this and adds a test case which fails without the fix applied. It also improves the control over which parameter is supposed to be ignored by providing a new method `protected boolean isParamToIgnore(java.lang.reflect.Parameter parameter)` which can be overridden by custom RequestBuilder implementations.


neiser:
Closing and re-opening PR as travis build failed after force push.

PR #113, opened 2019-10-11
Extend search for `@ApiResponse` annotations and merge everything together

neiser:
This PR extends how `@ApiResponse(s)` annotations are found when building the OpenAPI. 

I'd like to put this to discussions and I'm open for feedback on this one. It basically enables putting `@ApiResponse` annotations on abstract base controllers and as such having more "globally" defined responses for all methods. This is helpful (at least for me) in larger projects. If there's another way of doing things like this, please notify me and this PR could be closed.

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/AbstractResponseBuilder.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/AbstractResponseBuilder.java
@@ -10,6 +10,9 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import java.util.Optional;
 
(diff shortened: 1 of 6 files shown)


M0rfic:
I'm very much in favor of this CR/PR. Although the [open-api response docs](https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---Annotations#apiresponse) are a bit vague on this subject, not confirming or denying their usage at class/type level, I've successfully used them in a different project with JAX-RS (jersey) to define common responses (401, 500, etc), so theoretically at least, this should be fine as well.

On a separate note, I have opened issue #114 which is somewhat related to this one. In short, you should be able to define global response components, which you can then `ref` in your individual operation responses. It is not a direct replacement for the behavior which you are proposing (imho it's more like an addition to it and the already supported response definitions), but it can also be used to simplify things to a certain degree (if it's accepted and fixed).

springdoc:
Hi, 

 Great job. Proposal accepted!

PR #116, opened 2019-10-15
Improve response building, fix NPE

neiser:
After the search was extended for `@ApiResponse` in #113 I found that one should still add the response determined from the method if no explicit `@ApiResponse` annotation for this HTTP code is found. I think this improves the generated OpenApi spec especially when used for generating clients from it.

What do you think?

This PR also fixes a Null Pointer Exception encountered for a certain combination of `@ApiResponse` annotations.


springdoc:
Hi, 

 Your fix, and improvments of tests are now merged into master.
 For the code related to: Add response determined from method if not explicitly given by @ApiRe…
 It creates some regressions no existing tests, and will not be integrated for the moment.
 You can resubmit a new PR, if you have a better solution.

PR #117, opened 2019-10-16
Introduce magic response code constant

neiser:
In response to #116 this is my second attempt to improve on this.

I've added a "magic" constant which can be used with an `@ApiResponse` annotation to add a response code determined from the method. It will only do that if the response is not defined yet.

@springdoc what do you think?

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/AbstractResponseBuilder.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/AbstractResponseBuilder.java
@@ -126,9 +126,15 @@ public abstract class AbstractResponseBuilder {
 			methodAttributes.setWithApiResponseDoc(true);
 			for (io.swagger.v3.oas.annotations.responses.ApiResponse apiResponseAnnotation : apiResponseAnnotations) {
 				ApiResponse apiResponse = new ApiResponse();
+
+				String responseCodeFromAnnotation = getResponseCodeFromAnnotation(apiResponseAnnotation, method, isGeneric, apiResponseAnnotations, apiResponsesOp);
+				if(responseCodeFromAnnotation == null) {
+					continue;
+				}
+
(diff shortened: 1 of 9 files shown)


springdoc:
Hello, 

Some of your proposed changes have been integrated into the master.
Introducing this new constant is not a priority, as you can handle this case just using swagger-annotations, without any code change on springdoc-openapi.
If you don't need @ApiResponse on class level, you can just remove it, from the class definition.

neiser:
Hi @springdoc thanks for having a look. I would like to know how you would achieve the functionality I've demonstrated with the test case `HelloControllerWithGlobalApiResponse` provided in 06d1451f30508725218d15375778d9e026f2617a without adding explicit `@ApiResponse` annotations to every controller method (something which does not scale in larger codebases in my opinion).

The basic functionality I need to achieve is that, by default, the method signature should be investigated if there's no `@ApiResponse` matching the "default" response code of the method. The current code of springdoc-openapi only investigates the response code if no `@ApiResponse` annotations are present at all. This clashes with the new possibility from #113 to define "global" `@ApiResponse` annotations, for example defining shared error codes. 

I'd appreciate some example how to achieve this with the current implementation. Maybe I've just overlooked some code parts and/or misunderstood how to correctly put annotations.

dkroehan:
@springdoc Any update on the question @neiser posted?

springdoc:
Hi Guys, 

 I have added Test 53 for a sample code:
- https://github.com/springdoc/springdoc-openapi/tree/master/springdoc-openapi-webmvc-core/src/test/java/test/org/springdoc/api/app53

 You can note that the expected result is  exactly the same as the expected result by @neiser here (without introducing any new annotation):

- https://raw.githubusercontent.com/neiser/springdoc-openapi/06d1451f30508725218d15375778d9e026f2617a/springdoc-openapi-core/src/test/resources/results/app51.json

 If you see other cases not handled, you can add update the sample code with the actual/expected results, and we can discuss it.

PR #123, opened 2019-10-21
Able to translate Spring properties

GonFreaks:
The swagger annotation is able to translate a spring property place holder to the value present in a properties file.

an place holder example : ${test}

- if the key is present, the place holder is replace by the value
- if the key is not present, the place holder is show


springdoc:
Hello GonFreaks,

 Thank for your contribution. Very much appreciated.

 We didn't have people asking, for writing documentation, or reference properties in the application. yaml.

 There seems a lot of impact for this change. Especially on performance. 
 Ideally, i would have prefered simpler way to handle this translation of spring-propeties, without having to change all the implementation classes.

 Also, did you bench the impact before adding the feature and after?

GonFreaks:
Hello,

Thank you for the comment, for the performance impact, I launch the test before and after and it seems that it's a little impact with the feature (see attached files: 
[test performance before.txt](https://github.com/springdoc/springdoc-openapi/files/3798401/test.performance.before.txt)
[test performance after.txt](https://github.com/springdoc/springdoc-openapi/files/3798402/test.performance.after.txt)

Why I decided to propose the feature.

Actually, I work with springfox for swagger with springboot but I decided to try OpenApi with swagger and find your project. In springfox, they are the possibility to give a spring boot property for the swagger documentation.

Purpose of the feature:
clean the RestController class to avoid many line of documentation into the class.

For your purpose to have a simple way to implement the feature, I'm going to analyse and come back nearly to propose you a new solution if possible.

Thank you

PR #149, opened 2019-11-08
Open up for subclassing again + NPE fix

dawiking:
Previous PR to open base classes for subclassing outside of package was undone during recent refactorings.

Now added some subclasses in tests directory to make sure same thing doesn't happen again.

Also fixed an NPE in AbstractResponseBuilder when some endpoints have multiple media types and others don't.


springdoc:
Hi, 

 All the changes looks good.

PR #155, opened 2019-11-11
fix spring Pageable

aanno2:
springdoc-openapi maps the spring `Pageable` like any other scheme. However according to [1] this type needs special handling.

I have fixed that by implementing the ModelConverter `PageableSupportConverter` that will replace the 'real' type `org.springframework.data.domain.Pageable` with an appropriate replacement type.

[1] https://docs.spring.io/spring-data/rest/docs/3.2.1.RELEASE/reference/html/#paging-and-sorting

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/PageableSupportConverter.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/PageableSupportConverter.java
@@ -1,0 +1,30 @@
+package org.springdoc.core;
+
+import io.swagger.v3.core.converter.AnnotatedType;
+import io.swagger.v3.core.converter.ModelConverter;
+import io.swagger.v3.core.converter.ModelConverterContext;
+import io.swagger.v3.oas.models.media.Schema;
+import org.springdoc.replacements.Pageable;
+
+import java.util.Iterator;
+
+public class PageableSupportConverter implements ModelConverter {
+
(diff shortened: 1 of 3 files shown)


aanno2:
BTW, you could test that with app2 of https://github.com/springdoc/springdoc-openapi-demos/pull/2 

This is the relevant JSON with the fix:

```json
      "Pageable": {
        "required": [
          "page",
          "size"
        ],
        "type": "object",
        "properties": {
          "page": {
            "minimum": 0,
            "type": "integer",
            "format": "int32"
          },
          "size": {
            "maximum": 2000,
            "minimum": 1,
            "type": "integer",
            "format": "int32"
          },
          "sort": {
            "type": "array",
            "items": {
              "type": "string"
            }
          }
        }
      },
```

springdoc:
Hello, 

 Thank for your contribution. Your proposal works fine because you have introduced your custom implementation of Pageable.

 Additionally, the custom converter handles a class of spring-data-commons, which springdoc-openapi do not rely on.
We believe that implementation of custom converters should stay local to each project, in this case.

springdoc:
Finally, we have added the support of Pageable on a specefic module. #177

aanno2:
Just for the records:
* Is there a way to enable custom ModelConverters without modifying the sources (i.e. on CLI, gradle, mvn)?
* Thank you for implemention of #177. What does 'specific module' say? How to switch Pageable support on?

Issue #159, opened 2019-11-17
Add support for Kotlin Coroutines


joaodias14:
Kotlin coroutines add an extra `Continuation` parameter in the suspending function. This means that there is an additional parameter in our Controller's methods. Currently, we are facing two issues with this.

**1st Issue**
Depending on where the code is built, we get the following error while accessing Swagger:
> com.fasterxml.jackson.databind.JsonMappingException: Null key for a Map not allowed in JSON (use a converting NullKeySerializer?) (through reference chain: io.swagger.v3.oas.models.OpenAPI["paths"]->io.swagger.v3.oas.models.Paths["/srv/v1/{redisInstance}/keys/{key}"]->io.swagger.v3.oas.models.PathItem["put"]->io.swagger.v3.oas.models.Operation["requestBody"]->io.swagger.v3.oas.models.parameters.RequestBody["content"]->io.swagger.v3.oas.models.media.Content["*/*"]->io.swagger.v3.oas.models.media.MediaType["schema"]->io.swagger.v3.oas.models.media.ObjectSchema["properties"]->java.util.LinkedHashMap["null"])

We found that code built in my IDE works, but the one built on my colleague IDE or via command line does not work. We think the issue is related to the name that is given to the `Continuation` parameter. On the code that works, its name is `paramContinuation` and on the one that does not work, its name is `$completion`. Our assumption is that Springdoc/Swagger doesn't "like" this dollar sign in the name which leads to the exception above. You may be wondering if we've found the reason for this parameter name difference, but we didn't 😞 

**2nd Issue**
If we run the code built in my IDE (the one which actually works), we can access Swagger. However, we see that the `Continuation` parameter is now part of it in every endpoint with a suspending function. This parameter is considered a request body or query parameter depending on the HTTP method (our guess). We would like to have this parameter removed from the endpoints.

Our questions:
1. Do you have Kotlin Coroutines support on your roadmap (which may require ignoring `Continuation` parameters)?
2. If not (or until you do) is there any possibility to instruct Springdoc/Swagger to simply ignore a given class so that we can work-around this issue?

Thanks and keep up the good work!

springdoc:
Hi,

 Our library first goal is to cover the main functionalities using spring-boot/Java and swagger-annotations for OpenAPI 3.

 It's difficult to investigate on your issue to get the root cause, because the stackTrace you are providing is coming from Jackson.

[...]

PR #172, opened 2019-11-21
Issue #159: Support for Kotlin Coroutines added

joaodias14:
I've added support for Kotlin Coroutines by ignoring `Continuation` parameter (which is automatically added to a suspending function, aka controller method). I've decided to create an additional module so that only people using Kotlin Coroutines have additional code in their projects.
Additionally, I've changed the way you were getting the names of the parameters because using `LocalVariableTableParameterNameDiscoverer` was not working with the example I have. We already have the name of the parameters at that point via Parameters array, so I've decided to make it simpler. Was there any reason why you were using `LocalVariableTableParameterNameDiscoverer`?
I hope you find the code good enough to be added to the library!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -50,6 +50,7 @@
 
     <modules>
         <module>springdoc-openapi-common</module>
+        <module>springdoc-openapi-kotlin-coroutines</module>
         <module>springdoc-openapi-webmvc-core</module>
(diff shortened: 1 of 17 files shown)


joaodias14:
Thanks!

Issue #173, opened 2019-11-21
Is there a configuration for tag names?


chadmalla-ff:
For my project I just want to set the tag name to the name of the controller class. Is there a way where I can configure this in the `OpenAPI` bean so that I don't have to annotate my controller with `@Tag(name = "", description = "")` for each controller?

Thanks!

springdoc:
Hi, 
 
 
  You can add a custom implementation of RequestBuilder Bean at your application level, and then override customiseOperation.
 For example:
```java
@Override
protected Operation customiseOperation(Operation operation, HandlerMethod handlerMethod) {
	operation.addTagsItem(handlerMethod.getBeanType().getSimpleName());
	return operation;
}
```

springdoc:
Another PR proposed by @hisener  will be available on next release:1.2.10

PR #175, opened 2019-11-21
Generate a default tag name if no tag specified

hisener:
Relates to https://github.com/springdoc/springdoc-openapi/issues/173

One of the nice features of Springfox is that it generates default tags. So that we can see endpoints grouped by controller name. It is handy for applications with a bunch of controllers. I think springdoc should also provide a default tag name. WDYT?

Please note that I haven't fixed all of the tests.

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/OpenAPIBuilder.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/OpenAPIBuilder.java
@@ -1,5 +1,9 @@
 package org.springdoc.core;
 
+import static org.springdoc.core.Constants.DEFAULT_SERVER_DESCRIPTION;
+import static org.springdoc.core.Constants.DEFAULT_TITLE;
+import static org.springdoc.core.Constants.DEFAULT_VERSION;
+
 import io.swagger.v3.core.util.AnnotationsUtils;
 import io.swagger.v3.core.util.ReflectionUtils;
 import io.swagger.v3.oas.annotations.Hidden;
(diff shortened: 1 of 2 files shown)


springdoc-openapi-common/src/main/java/org/springdoc/core/OpenAPIBuilder.java
```diff
@@ -94,14 +94,14 @@ else if (openAPI.getInfo() == null) {
...
+                ReflectionUtils.getRepeatableAnnotations(handlerMethod.getBeanType(), Tag.class);
 
         // method tags
-        List<io.swagger.v3.oas.annotations.tags.Tag> methodTags = ReflectionUtils
-                .getRepeatableAnnotations(handlerMethod.getMethod(), io.swagger.v3.oas.annotations.tags.Tag.class);
+        List<Tag> methodTags =
+                ReflectionUtils.getRepeatableAnnotations(handlerMethod.getMethod(), Tag.class);
 
-        List<io.swagger.v3.oas.annotations.tags.Tag> allTags = new ArrayList<>();
+        List<Tag> allTags = new ArrayList<>();
```
hisener: It's unrelated to the change but unnecessary qualification.

chadmalla-ff:
Hi @hisener  is the purpose of this PR to have a default tag name to be the controller name without annotating the controller with`@Tag`? Thank you for the PR btw!

chadmalla-ff:
Never mind just tried with version `1.2.10` and doing exactly what I wanted in my original issue. Thanks!

PR #208, opened 2019-12-02
Add Spring Boot metadata for config properties.

marblenix:
On compile, generate the spring-configuration-metadata.json described in the [configuration-metadata](https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-configuration-metadata.html#configuration-metadata-annotation-processor) documentation.

This is helpful for IDEs to autocomplete and give documentation for properties in the application.properties file.

### DIFF
--- a/springdoc-openapi-common/pom.xml
+++ b/springdoc-openapi-common/pom.xml
@@ -13,6 +13,11 @@
             <artifactId>spring-boot-autoconfigure</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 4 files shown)


marblenix:
This is a very basic first pass. The actual instance of the class created by Spring is not used in any way, as I opted to let the Constants class and the `@Values` currently defined do it's own thing. A more complete pull request would be to refactor those value annotations to use the instance of SpringDocConfigProperties created on startup. Though since I just wanted the documentation I decided to leave those out.

If it's okay to start working on that I would like to make a new pull request for refactoring out those `@Values` in the next week or two.

springdoc:

PR #226, opened 2019-12-05
fix link

natrem:


springdoc:
Thank you for your contribution @natrem

PR #254, opened 2019-12-11
fix deprecation

natrem:
fix deprecation of WebMvcConfigurerAdapter


natrem:
my bad.
proposal is pointless then.

PR #255, opened 2019-12-11
Added ability to ignore param with @Hidden annotation at class level

marronef:


springdoc:
thank you for your contribution @marronef

PR #281, opened 2019-12-18
make validatorUrl configurable and fix overwrites 

### LINKED ISSUE #278: Allow proper configuration of validatorUrl and configUrl
I looked at [SwaggerWelcome](https://github.com/springdoc/springdoc-openapi/blob/master/springdoc-openapi-ui/src/main/java/org/springdoc/ui/SwaggerWelcome.java) which redirects to the included `swagger-ui` adding configured queryParameters.

The `url` and `validatorUrl` parameters are always set.
That is why it is currently not possible to specify a `validatorUrl`, or set the `url` property in a config provided by a config-endpoint (`configUrl`).
This is because of the [precedence of configuration](https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/configuration.md) properties for swagger-ui: 

```
From lowest to highest precedence:

* The swagger-config.yaml in the project root directory, if it exists, is baked into the application
* configuration object passed as an argument to Swagger UI (SwaggerUI({ ... }))
* configuration document fetched from a specified configUrl
* configuration items passed as key/value pairs in the URL query string
```

It would be nice to 
* be able to specify the `validatorUrl`
* have proper support for `configUrl`

(It is possible to get around this by disabling SwaggerConfig / SwaggerWelcome + providing custom redirects)

kaibra:
fixes #278

This PR introduces support to configure a `validatorUrl` for swager-ui.
Also fixed:
In case `configUrl` is used, `url`, which is derived from springdoc configs is not added anymore as well as the empty default for validatorUrl. 
Both would overwrite entries in the configured 'configUrl' endpoint.

Regards,

Kai

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/Constants.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/Constants.java
@@ -18,8 +18,8 @@ public final class Constants {
     public static final String SPRINGDOC_ACTUATOR_TAG = "Actuator";
     public static final String DEFAULT_WEB_JARS_PREFIX_URL = "/webjars";
     public static final String WEB_JARS_PREFIX_URL = "${springdoc.webjars.prefix:#{T(org.springdoc.core.Constants).DEFAULT_WEB_JARS_PREFIX_URL}}";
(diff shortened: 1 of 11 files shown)


springdoc:
Thank you @kaibra  for your contribution.

PR #293, opened 2019-12-23
SPI for customizing operations, parameters and properties

m-kay:
make customizing operations, parameters and properties available through an SPI


m-kay:
@springdoc could you may have a look at this? Checks seem to fail because of some wrong nexus configuration, but I could run all tests locally.

bnasslahsen:
It will be available on the next release of springdoc-openapi: v1.2.22

PR #295, opened 2019-12-24
GroupedOpenApi: replace array parameter with vararg

Aloren:
This change is backward compatible, but allows more convenient usage of `GroupedOpenApi` api.
Instead of:
```
	String paths[] = {"/store/**", "/admin/**"};
	return GroupedOpenApi.builder()
                   .setGroup("stores")
                   .pathsToMatch(paths)
		  .build();
```
now you can use:
```
	return GroupedOpenApi.builder()
                   .setGroup("stores")
                   .pathsToMatch("/store/**", "/admin/**")
		  .build();
```


springdoc:
Thank you @Aloren for your contribution!

PR #296, opened 2019-12-24
Property `springdoc.api-docs.groups.enabled` removed

Aloren:
As not needed any more for enabling multiple OpenAPI definitions support.

Adding beans of type `GroupedOpenApi` should be enough for `springdoc` library to enable multiple OpenAPI definitions support.

Also I've found that currently endpoint `/swagger-config` for ` springdoc-openapi-webflux-ui` returns incorrect `urls` for multiple OpenAPI definitions. Should I fix this in this PR or create separate one?


springdoc:
Hi, 

 For the endpoint /swagger-config on springdoc-openapi-webflux-ui whcih returns incorrect urls for multiple OpenAPI definitions, its better to create a different PR.

PR #299, opened 2019-12-26
Fix autoconfigurations

Aloren:
Fix autoconfigurations to be not loaded if classpath does not contain mvc/reactive dependencies or application is not web application.

Also had to move 2 beans from `SpringDocConfiguration` to `SpringDocWebMvcConfiguration` and `SpringDocWebFluxConfiguration`, as `SpringDocWebMvcConfiguration`/`SpringDocWebFluxConfiguration` could not be disabled, they had circular reference, so it was not possible to disable any of the configuration as it caused issues with bean creation.


springdoc-openapi-common/src/main/java/org/springdoc/core/SpringDocConfiguration.java
```diff
@@ -39,23 +39,11 @@ public OpenAPIBuilder openAPIBuilder(Optional<OpenAPI> openAPI, ApplicationConte
         return new OpenAPIBuilder(openAPI, context, securityParser,modelConverters);
     }
 
-    @Bean
```
Aloren: Unfortunately had to move creation of these beans to `SpringDocWebMvcConfiguration`+`SpringDocWebFluxConfiguration`, as it was not possible to disable the latter

springdoc-openapi-webflux-core/src/main/java/org/springdoc/core/SpringDocWebFluxConfiguration.java
```diff
@@ -2,18 +2,23 @@
...
+import reactor.core.publisher.Mono;
 
 import java.util.List;
 import java.util.Optional;
 
 import static org.springdoc.core.Constants.SPRINGDOC_ENABLED;
 
 @Configuration
+@ConditionalOnWebApplication(type = ConditionalOnWebApplication.Type.REACTIVE)
+@ConditionalOnClass(Mono.class)
```
Aloren: We need to check both reactive dependency is present and application is reactive web aaplication

springdoc-openapi-webmvc-core/src/main/java/org/springdoc/core/MultipleOpenApiSupportConfiguration.java
```diff
@@ -18,6 +21,8 @@
 
 
 @Configuration
+@ConditionalOnWebApplication(type = ConditionalOnWebApplication.Type.SERVLET)
+@ConditionalOnClass(DispatcherServlet.class)
```
Aloren: We need to check both mvc dependency is present and application is blocking web application

springdoc-openapi-webmvc-core/src/main/java/org/springdoc/core/SpringDocWebMvcActuatorConfiguration.java
```diff
@@ -1,20 +0,0 @@
...
-import org.springframework.boot.actuate.endpoint.web.servlet.WebMvcEndpointHandlerMapping;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
-import org.springframework.context.annotation.Bean;
-import org.springframework.context.annotation.Configuration;
-
-import static org.springdoc.core.Constants.SPRINGDOC_SHOW_ACTUATOR;
-
-@Configuration
-@ConditionalOnProperty(name = SPRINGDOC_SHOW_ACTUATOR)
-public class SpringDocWebMvcActuatorConfiguration {
```
Aloren: Moved to `SpringDocWebMvcConfiguration` as internal class, so that we do not need to duplicate all conditions for checking mvc stack

springdoc-openapi-webmvc-core/src/main/java/org/springdoc/core/SpringDocWebMvcConfiguration.java
```diff
@@ -46,4 +65,16 @@ public ResponseBuilder responseBuilder(OperationBuilder operationBuilder) {
         return new ResponseBuilder(operationBuilder);
     }
 
+    @Configuration
+    @ConditionalOnProperty(name = SPRINGDOC_SHOW_ACTUATOR)
+    @ConditionalOnClass(WebMvcEndpointHandlerMapping.class)
```
Aloren: Added in case somebody forgets to add actuator dependency

bnasslahsen:
Thank you @Aloren for your contribution.

PR #313, opened 2020-01-02
fix wrong server url when goruped api name contains encoded character

### LINKED ISSUE #312: Wrong server url with when grouped api name contains special charater
When the name of the grouped API contains a character which gets URL encoded the server url is generated wrong. When calculating the server URL the request URL should be decoded.

m-kay:
fixes #312


m-kay:
Besides the fix I also added a configuration to make the server url protocol relative which is needed when there is a component before the service making the ssl termination.

bnasslahsen:
Thanks @m-kay for your contribution

bnasslahsen:
For the ssl termination, if the application is hosted behind a proxy. There is no need for additionnal changes on springdoc.
We can rely on native spring-boot configuration.
You can have a look at the related settings:

m-kay:
@bnasslahsen thanks for the hint, that did the trick.

Should I submit a new pull request without the relativ url configuration or do you think this could still be useful for someone? At least the fix with injecting the properties via constructor should be included because otherwise the `showActuator` configuration won't work when grouped apis are used.

bnasslahsen:
Hi @m-kay, 

 Yes you can update your PR. 
 Thank you.

m-kay:
@bnasslahsen I have now removed the relative baseurl configuration.

bnasslahsen:
It will be available on the next release: v1.2.23

PR #372, opened 2020-01-23
Support for CompletionStage as endpoint return type. Fixes #370.

### LINKED ISSUE #370: Add support for async mappings returning CompletableFuture.
Spring supports returning CompletableFutures for mappings marked with @Async. Should be pretty similar to https://github.com/springdoc/springdoc-openapi/issues/297.

mvh77:


springdoc-openapi-webmvc-core/src/main/java/org/springdoc/core/ResponseBuilder.java
```diff
@@ -56,4 +61,12 @@ else if (Callable.class.getName().contentEquals(parameterizedType.getRawType().g
...
 	}
 
+	private boolean isParameterizedTypeWrapper(Class<?> clazz, ParameterizedType type) {
+		try {
+			Class<?> parameterType = Class.forName(type.getRawType().getTypeName());
+			return clazz.isAssignableFrom(parameterType);
+		} catch (ClassNotFoundException e) {
+			return false;
+		}
+	}
```
mvh77: we want to support any return type inherited from CompletionStage, like ComplatableFuture

bnasslahsen:
Thank you @mvh77 for your contribution. 
It has been merged with other pending changes and will be available on the next release v1.2.29

PR #375, opened 2020-01-24
Support java 8 Optional for query parameter

WellingR:
Spring automatically treats query parameters as optional if they are wrapped in `Optional<T>` however I notices that this was not shown in the api docs.

This change ensures that a @RequestParam is marked not required if the type of the parameter is wrapped in Optional


bnasslahsen:
Thank you @WellingR for your contribution.
Your PR is merged with master and will be available on the next release: v1.2.29

PR #393, opened 2020-01-30
fix: prevent swagger-ui default validation

### LINKED ISSUE #394: default validation is active
Currently the default validation for swagger-ui is enabled if you do not specify an explicit validatorUrl which is not null and also not "" (empty string).

not sure if this is only a problem when using groups?!

kaibra:
Not providing the validatorUrl currently enables the default validation for swagger.
This brings back the better default.
Not sure if this is the right place to fix it

fixes #394


bnasslahsen:

m-kay:
Thank you, but credits should go to @kaibra 😄

bnasslahsen:
Yeah right @m-kay. 
My apologies for this mispelling @kaibra.

PR #411, opened 2020-02-11
Update pom.xml

sawant-abhijeet:


bnasslahsen:
Thank you @sawant-abhijeet for your contribution

Issue #269, opened 2019-12-14
Kotlin Support ?


DavidRomao:
When generating the documentation It is not marking non nullable types as required. As I was not able to any information if kotlin is supported, I would like to know if kotlin is not supported or am I missing something? 

I could not get the fields to show up as required with @Nonnull or @Schema(required=true) as well.

joaodias14:
Kotlin is supported, but to my knowledge, there is no special rule for non-nullable Kotlin types. This means that `@RequestParam` is the one influencing the fact that a given variable is required or not.
The following works as expected:
`suspend fun keys(@PathVariable redisInstance: RedisInstance, @RequestParam(required = false) test: String?)`

DavidRomao:
I am actually referring to the object used with @RequestBody, not a paremeter. Sorry I should have been more specific.  I wanted to set the field of a class as required. 

I expected this to work but It didn't mark the field as required.
`data class RegisterUserDTO(
                           @Schema(required=true)
                           val email: String,
                           @Schema(required=true)
                           val password: String,
                           @Schema(required=true)
                           val name: String
)
fun register(@RequestBody user : RegisterUserDTO )
`

joaodias14:
Oh sorry, I thought you were talking about query parameters.
To make that work, you must use `@field:Schema` instead of `@Schema`. You must use this `field:` prefix if you define classes via constructor declaration and want to apply a given annotation to one of the attributes. Take a look at https://kotlinlang.org/docs/reference/annotations.html#annotation-use-site-targets.

springdoc:
Thank you @joaodias14  for the support.

WtfJoke:
Hi there

A follow up question, we have kind of a similar issue.

We have following classes:

```kotlin
data class RegistrationForm(
    @field:Schema(required = true) val businessName: String,
    @field:Schema(required = true) val address: Address,
)

data class Address(
    @field:Schema(required = true) val street: String,
    @field:Schema(required = true) val streetNumber: String,
    @field:Schema(required = true) val postCode: String,
    @field:Schema(required = true) val city: String
)
```

Setting the required property works for all fields, except for the address field - see the attached screenshot of the swagger ui
![swaggerregistrationform](https://user-images.githubusercontent.com/7139697/71181068-b581fd80-2273-11ea-8680-a68424c372d3.PNG)

Do you have an idea what could cause this and how to fix it? Thanks in advance

joaodias14:
[...]

PR #413, opened 2020-02-11
Add support for @NotNull annotations emitted via Kotlin compiler

rohanprabhu-jm:
The kotlin compiler does not emit JSR-305 annotations, which is what `AbstractRequestBuilder` looks for when marking any specific type/parameter as "required". This causes kotlin developers to additionally add a `@NotNull` annotation to each property (from the JSR package), as the kotlin compiler emits `org.jetbrains.annotations.NotNull`.

This PR adds the kotlin specific annotation to the list of annotations that qualify a type/param to be marked as "required" in a schema.


OliverGrack:
Is there any information on why this Pull request got closed?

goodhoko:
> Is there any information on why this Pull request got closed?

@OliverGrack see https://github.com/springdoc/springdoc-openapi/issues/269#issuecomment-584607858

PR #416, opened 2020-02-12
Migrate to Jakarta from EE and update POM versions

nathankooij:
In Java 11 EE was dropped from the JDK, after which the dependencies were subsequently moved under the [Jakarta umbrella](https://jakarta.ee/). These are (currently) exact duplicates of the original `javax.*` dependencies, except renamed. To avoid conflicts with (newer) downstream projects we should also apply the Jakarta dependencies.

While there, I also fixed the build as the parent POM's version had drifted from the children POMs.


bnasslahsen:
Thank you @nathankooij for your contribution.

PR #427, opened 2020-02-14
Fixes #418 by adjusting the output for Links.

### LINKED ISSUE #418: Fix support for spring's HATEOAS data serialization transformations and provide FAQ example complex transformation
Spring HATEOAS uses custom serializers for the couple of specific data types it uses, and springdoc doesn't know about these data types and thus maps them incorrectly to swagger.

Ideally there would be support for the automatic correct transformation of these data types in springdoc, otherwise, an example of how to use a PropertyCustomizer or similar to perform such mappings would be very helpful.

In particular:

The org.springframework.hateoas.RepresentationModel class represents a resource's links as a Collection<> named 'content' which needs to be translated in JSON to a field named '_embedded' of the same type.

The org.springframework.hateoas.CollectionModel class represents a resource's embedded contents as a List<Link> named 'links' which needs to be translated in JSON to a field named '_links' of type Map<String, Link> 

I tried naively implementing a `PropertyCustomizer` to do this transformation but this didn't work, and there doesn't seem to be an example of a non-trivial PropertyCustomizer to base work from. Below is my failed attempt.

Ideally support for these types would be built-in to springdoc, otherwise a FAQ on how one might implement a suitable customizer would be even more flexible as it would allow others to implement other non-trivial transformations.

Lastly, many thanks for this project, it looks to be very high quality, excellent job!

```
@Component
public class OpenApiHateosFixup implements PropertyCustomizer {
    @Override
    public Schema customize(Schema property, AnnotatedType type) {

        // map from List<Link> to example Map<String,Link>
        if (property instanceof ArraySchema) {
            String typeRef = ((ArraySchema) property).getItems().get$ref();
            if (typeRef != null && typeRef.equals("#/components/schemas/Link")) {
                property = new MapSchema().type(typeRef); // seems to be ignored
            }
        }
        if (property != null) {
            // HATEOAS: map from 'links' field to '_links' and from "content" to "_embedded"
            if ("links".equals(type.getPropertyName())) {
                property.name("_links");  // seems to be ignored
            } else if ("content".equals(type.getPropertyName())) {
                property.name("_embedded");  // seems to be ignored
            }
        }
        return property;
    }

}
```

fredrik-w:
There is a custom serializer for the Links object which creates a sorted map of all links. That map is then passed to jackson for serialization, thus some manual work is needed to fix the schema.

### DIFF
--- a/springdoc-openapi-data-rest/src/main/java/org/springdoc/core/SpringDocDataRestConfiguration.java
+++ b/springdoc-openapi-data-rest/src/main/java/org/springdoc/core/SpringDocDataRestConfiguration.java
@@ -18,12 +18,26 @@
 
 package org.springdoc.core;
 
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import io.swagger.v3.core.converter.AnnotatedType;
+import io.swagger.v3.core.converter.ModelConverters;
+import io.swagger.v3.core.converter.ResolvedSchema;
+import io.swagger.v3.core.util.Json;
+import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.models.media.MapSchema;
+import io.swagger.v3.oas.models.media.ObjectSchema;
+import io.swagger.v3.oas.models.media.StringSchema;
 import org.springdoc.core.converters.PageableSupportConverter;
-
+import org.springdoc.core.customizers.OpenApiCustomiser;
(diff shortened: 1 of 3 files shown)


fredrik-w:
Is this the way to go, to attach directly to `OpenAPI`, or should we take some other route?

fredrik-w:
Reworked the implementation to use an `OpenApiCustomiser` instead, which feels more inline with the project.

bnasslahsen:
Thank you @fredrik-w for your contribution.

PR #450, opened 2020-02-19
Add missing @ConditionalOnMissingBean annotations

ptahchiev:
Add missing @ConditionalOnMissingBean annotations  to allow users to override the bean


ptahchiev:
@bnasslahsen done

PR #451, opened 2020-02-19
Add OpenApiBuilderCustomiser

ptahchiev:
Add OpenApiBuilderCustomiser to allow users to customize the OpenApiBuilder.
In our case we have our own custom annotation `FacadeRestResource` (much similar to spring-data-rest `RepositoryRestResource`) and we need a way to allow the OpenApiBuilder to scan beans for our annotation.
With the proposed change we introduce a new interface called `OpenApiBuilderCustomiser` and thus we can create our own customizer to provide the additional Map of mappings.


ptahchiev:
Hi @bnasslahsen - I've added a test-case.

ptahchiev:
@bnasslahsen  Is there anything else I need to improve on the PR? The longer it waits the more conflicts appear - i see now you added 92 test and now I have conflicts with my branch :(

bnasslahsen:
Hi @ptahchiev, 

 Your PR is merged with master and will be available on the next release: 1.2.33

PR #487, opened 2020-03-11
Delegate parameter primitive type generation to swagger-core

nathankooij:
Originally we derived the schema ourselves when a primitive type was detected. The underlying `swagger-core` library (and in specific the `ModelResolver`) also does this for us, thus making this piece of code superfluous. Additionally this helps converge schema generation behavior for different generation paths (parameters, response bodies, etc.). A benefit of this is that any additional configuration like custom `ModelConverter`s are now applied.

**Some extra background**
We have configured a custom `ModelConverter` that can rewrite certain types for us on the fly. This works well for responses, but we noticed that this did not work for parameters that were primitive types, as they bypassed the normal code path for schema generation. One such example is an `Instant`, which Swagger expects to be a `long`, but which we have configured to be a date time string. So our `ModelConverter` correctly transforms the type for model properties, but not parameters which is a bit unexpected. This PR aims to correct that behavior.

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/SpringDocAnnotationsUtils.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/SpringDocAnnotationsUtils.java
@@ -42,14 +42,7 @@ public class SpringDocAnnotationsUtils extends AnnotationsUtils {
 
 	public static Schema resolveSchemaFromType(Class<?> schemaImplementation, Components components,
 			JsonView jsonView) {
-		Schema schemaObject;
-		PrimitiveType primitiveType = PrimitiveType.fromType(schemaImplementation);
-		if (primitiveType != null) {
-			schemaObject = primitiveType.createProperty();
-		}
-		else {
-			schemaObject = extractSchema(components, schemaImplementation, jsonView);
-		}
+		Schema schemaObject = extractSchema(components, schemaImplementation, jsonView);
 		if (schemaObject != null && StringUtils.isBlank(schemaObject.get$ref())
 				&& StringUtils.isBlank(schemaObject.getType())) {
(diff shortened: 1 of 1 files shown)


nathankooij:
Hi @bnasslahsen,

I had a look earlier today. I noticed that the previous commit (or basically what is on `master` right now) also failed for me. However, the failures were not consistent. Sometimes the tests would pass, and sometimes they would not. But if they did fail, it was always in the `springdoc-openapi-kotlin` module.

I captured output from the `SpringDocApp1Test` in that module: once when the test passed, and once when it failed. No code between test runs was changed. Please find this output attached:

[app1_fails.txt](https://github.com/springdoc/springdoc-openapi/files/4330436/app1_fails.txt)
[app1_passes.txt](https://github.com/springdoc/springdoc-openapi/files/4330437/app1_passes.txt)

I believe in cases where it fails is due to this:
```
   SpringDocWebFluxConfiguration#openApiResource:
      Did not match:
         - @ConditionalOnMissingBean (types: org.springdoc.api.OpenApiResource; SearchStrategy: all) found beans of type 'org.springdoc.api.OpenApiResource' openApiResource (OnBeanCondition)
```
as that would explain the `404` later on as this bean serves the endpoints being called in the test. However, I have had the test fail in a similar fashion even with an `@Autowired OpenApiResource resouce` in test, where `(resource == null) = true`, yet the bean was not created for exactly the same reason.

I'm happy to take another look later, but I'm a bit out of ideas. I was wondering if you had any insights what might cause this peculiar behavior?

bnasslahsen:
@nathankooij, 

 I have merged your changes with master. 
 I don't think they are the cause of CI build failing.
 And right now, build is passing on master.

PR #490, opened 2020-03-15
Implemented QDSLPredicate support using operation customizer

gibahjoe:
This implementation uses the operation customizer to replace the Predicate with the fields in the QueryDsl bindings which then appear as query parameter fields as opposed to being an object like the previous implementation.

I left the earlier one and used a config parameter to check for preferred implementation to apply. It defaults to this implementation though

Latest


bnasslahsen:
Hi @gibahjoe , 

 I find the second implementation much better, as it doesn't rely on javassist.

 For the test SpringDocApp98Test, it should pass even without using the property `springdoc.qdslpredicate.mode`. When using the second implemntation But its not the case. Is it normal ?

gibahjoe:
Thanks. I thought as much.

The test won't pass because without the springdoc.qdslpredicate.mode, the default implementation (this new implementation) takes effect which removes the initial predicate parameter from the operation.

bnasslahsen:
Hi @gibahjoe, 

We shouldn't keep many implementations of the same functionality.
I will keep your second proposal, which is the default.

 If you have any reasons, why you would like to keep the old implemntation, please don't hesitate to share.

bnasslahsen:
@gibahjoe, 

 The generated OpenAPI description contains the following fileds: notStatuses and child. 
 These fields, should not be generated on the OpenAPI description, because the  DummyEntityPredicate doesn't declare them.

gibahjoe:
> Hi @gibahjoe,
> 
> We shouldn't keep many implementations of the same functionality.
> I will keep your second proposal, which is the default.
> 
> If you have any reasons, why you would like to keep the old implemntation, please don't hesitate to share.

Imagine a predicate object that has 10 fields. When you run the spec file generated from the first proposal through an openapi client sdk generator, it generates a method with an object param that has 10 fields right? but if you run the spec generated by the second proposal through the same generator, it will generate a method with 10 optional params.

This might not mean much to dart clients since dart has named arguments but for typescript for instance, thats a long ass method.

Basically
`searchUsers(name?: string, firstName?: string, lastName?: string, middleName?: string, job?: string, age?: string)`
versus
`searchUsers(filter)` where filter is an object that declares the fields above.

so it then comes down to user preference. Thats why i think it's best to keep both and make it configurable.

But whatever decision you make, its fine.

gibahjoe:
> The generated OpenAPI description contains the following fileds: notStatuses and child.
> These fields, should not be generated on the OpenAPI description, because the DummyEntityPredicate doesn't declare them.

By default, `@QueryDslPredicate` maps every field in an Entity (the root) to a query param of the same name. The `DummyEntityPredicate` is used to define custom mappings or override a mapping. 

In this case,  `code` was given an alias `notCode` so that means `notCode` overrides `code` in the query params but the rest of the fields not declared in the  predicate stays the same.

Thats how it works. (Declared fields - overriden fields) + binding aliases.

bnasslahsen:
Thank you @gibahjoe for your feeback.
 You didn't mention the usage of `QDummyEntity`. Which means, explicit bindings are configured and will be used.

 `QDummyEntity `, doesn't contain the child and notStatuses fields.  Therefore, this fields shouldn't appear in OpenAPI description. 

 If you try the test from the UI, and you fill `child `or `notStatuses `fields, you will get the following error:
`java.lang.IllegalArgumentException: Unable to find field notStatuses on class test.org.springdoc.api.QDummyEntity!`

 This means, the generated spec is not correct.

gibahjoe:
> Thank you @gibahjoe for your feeback.
> You didn't mention the usage of `QDummyEntity`. Which means, explicit bindings are configured and will be used.
> 
> `QDummyEntity `, doesn't contain the child and notStatuses fields. Therefore, this fields shouldn't appear in OpenAPI description.
> 
> If you try the test from the UI, and you fill `child `or `notStatuses `fields, you will get the following error:
> `java.lang.IllegalArgumentException: Unable to find field notStatuses on class test.org.springdoc.api.QDummyEntity!`
> 
> This means, the generated spec is not correct.

Okay. I understand.

Usually, QDummyEntity is generated  by querydsl annotation processor and will always have the all the fields in DummyEntity. But because I created `QDummyEntity` manually, I forgot to add those missing fields, thats why its complaining. I will add them now and push up.

gibahjoe:
I see you reverted?

bnasslahsen:
> I see you reverted and removed the config files?

Just temporarily from master for the release.
You can use the following branch to push your changes.

I will merge your PR soon after.

gibahjoe:
> > I see you reverted and removed the config files?
> 
> Just temporarily from master for the release.
> You can use the following branch to push your changes: qdsl-predicate-feature
> 
> I will merge your PR soon after.

oops. already updated PR

PR #492, opened 2020-03-16
Permit to enforce calculated server scheme to run on a https proxy

liuyx11241:
add a new property `springdoc.calculated-server-scheme` to enforce overriding the calculated server scheme. In case of running a http swagger on a https proxy

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/SpringDocConfigProperties.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/SpringDocConfigProperties.java
@@ -61,6 +61,11 @@ public class SpringDocConfigProperties {
 
 	private String defaultProducesMediaType = MediaType.ALL_VALUE;
 
+	/**
+	 * To override calculated server scheme in order to permit run on a https proxy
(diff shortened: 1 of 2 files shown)


codesorcery:
I'm currently preparing a PR for an implementation which automatically takes the _Forwarded_ header and the _X-Forwarded-*_ headers into account. Might be an alternative to providing a hardcoded value.

bnasslahsen:
Hi @liuyx11241 and @codesorcery, 

 There is no problem if your application uses **https**.

If your application is deployed behind a proxy who sets the https connection. You can configure the spring-boot application to use the url of the proxy: There are relevant spring-boot properties for headers forwarding depending on your version.

Or you can customize OpenAPI object programmatically and set the server url. But this should be the last solution for the issue you are facing

- https://springdoc.github.io/springdoc-openapi-demos/faq.html#can-i-customize-openapi-object-programmatically

codesorcery:
Hey @bnasslahsen,
thanks for the fast reply! Somehow I must have overlooked that part when I read the FAQ.

PR #494, opened 2020-03-17
Use Forwarded and X-Forwarded-* headers when calculating server URL

codesorcery:
When the application is served behind a proxy (e.g. for load balancing),
the proxy can provide the de-facto standard X-Forwarded-Proto, X-Forwarded-Host
and X-Forwarded-Port headers to submit the protocol, host and port of the original
request to the downstream application.

Using Spring's UriComponentsBuilder we can use this information to construct
the server URL of the original request.


bnasslahsen:
@codesorcery, 
 
 I have just added my comment on the PR #492.

 You just need to use the right spring-boot properties to use the X-forwarded headers. Spring-boot handles this correctly.

Please have a look at our FAQ: 

PR #500, opened 2020-03-18
use MethodParameter#isOptional() instead of checking for Optional type

zarebski-m:
Check for whether a parameter is required now uses `isOptional()` method. This improves integration with JSR-305 annotations and Kotlin nullability.

In the following Kotlin example, parameter `foo` will be mandatory, but `bar` and `baz` not:
```kotlin
@RestController
@RequestMapping
class FooController {
    @GetMapping("test")
    fun testParams(
        @RequestParam foo: String,
        @RequestParam bar: String = "default",
        @RequestParam baz: String?
    ) {}
}
```

Similarly, parameter `foo` will correctly be rendered non-mandatory in this Java example:
```java
@RestController
@RequestMapping
public class FooController {
    @GetMapping("test")
    void fooNullable(@RequestParam @Nullable String foo) {}
}
```


bnasslahsen:

PR #501, opened 2020-03-20
add property resolver on @Schema.description

tdevilleduc:


bnasslahsen:
Thank you @tdevilleduc  for your contribution.
Your implementation will be available through the following property: 

```properties
springdoc.api-docs.resolve-schema-properties=true
```

Issue #268, opened 2019-12-14
Request parameter object is shown incorrectly on UI


martin-tarjanyi:
Request parameter object of GET request is shown as JSON on swagger-ui instead of individual parameters.

![image](https://user-images.githubusercontent.com/17810336/70853109-f364d780-1ea9-11ea-8e24-f98b49875add.png)

This issue has been mentioned a few times before:
* https://github.com/springdoc/springdoc-openapi/issues/120
* https://github.com/springdoc/springdoc-openapi/issues/67
* https://github.com/springdoc/springdoc-openapi/issues/12

These all have been closed as they are caused by an [issue](https://github.com/swagger-api/swagger-ui/issues/4581) in swagger-ui project. However, that issue was opened 18 months ago and there is not much activity on that project so I don't expect it to be fixed anytime soon (if ever).

Would it be possible to generate parameters in the OpenAPI docs such a way that it can be correctly presented by swagger-ui?

I created a demo project with 2 endpoints:
https://github.com/martin-tarjanyi/openapi-ui-issue/blob/master/src/main/java/com/example/openapiuiissue/DemoController.java#L13-L23

As I understand semantically there is no difference between them. However, the generated API docs (and in consequence the UI) is quite different.

I'm rising this issue again because this is the only thing which prevents me from moving all my projects to springdoc. I also would be happy to contribute to solve this issue if possible.

springdoc:
Hi @martin-tarjanyi, 

 You can achieve the same result using the swagger-annotations:

```java
@Parameters({@Parameter(in = ParameterIn.QUERY, name = "name", schema = @Schema(type = "string"), required = true),
		@Parameter(in = ParameterIn.QUERY, name = "age", schema = @Schema(type = "integer", format = "int32"), required = true)})
@GetMapping("1")
public String demo(@Parameter(hidden = true) MyParameterObject parameterObject) {
	return "demo";
}
```

If you have a better proposal, feel free to contribute.

Please don't open many issues for the same topic. You can add your comment, even the issue is closed and we will reopen it, if needed.

daniel-frak:
[...]

Issue #162, opened 2019-11-18
Any plan to make the POJO parameter binding same as SpringMVC ?


JaynLau:
Now, the POJO object without @RequestBody is binded to request body, but in SpringMVC, the properties of the POJO object without @RequestBody are binded to request parameters.

Is there any plan to change this same as SpringMVC ?

springdoc:
Hi, 

This behaviour depends on the HTTP method you are using.
For example, if you use `@GetMapping` the POJO object without `@Requestbody` will be bound to request parameters.
If you use `@PostMapping` the POJO object without `@Requestbody` will be bound to request body in order to be compliant with what the spring-rest-mvc API is expecting.
 
 You can add sample Java Code with the expected and the actual OpenAPI description (Json/yml) , if you think there any potential issue.

quaff:
> If you use @PostMapping the POJO object without @RequestBody will be bound to request body in order to be compliant with what the spring-rest-mvc API is expecting.

@bnasslahsen I'm afraid you are wrong, POJO without `@RequestBody` will always be bound from request parameters.
According to that, `@ParameterObject` introduced by #505 is redundant since all unannotated POJO should be treated as parameter object, as springfox does.

Issue #119, opened 2019-10-16
Add support for @ModelAttribute annotations


ckubik-monster:
It would be nice to have support also for @ModelAttribute annotation used in a parameter of controller method.

Example:
```java
@RestController
public class ExampleController {

  @GetMapping("/example")
  void get(@ModelAttribute InputDTO input) {
    //...
  }
}

class InputDTO {
  @Schema(required = true, description = "...")
  String first;

  @Schema(description = "...")
  String second;
}
```

Now is such parameter processed with no special care what leads to false API docs.

For Java example above it would lead to:
```javascript
//parameters for path /example
"parameters": [
  {
    "name":"input",
    "in":"query",
    "required":true,
    "schema":{"$ref":"#/components/schemas/InputDTO"}
  }
]

//schema for InputDTO
{
  "required":["first"],
  "type": "object",
  "properties": {
    "first": {
      "type": "string",
      "description" :"..."
    },
    "second": {
      "type": "string",
      "description": "..."
    }
  }
}
```

As a result Spring will handle request parameters in different way than what API doc is declaring and also Swagger UI is unable to call a such endpoint of course.

Expected API docs for the example is following:
```javascript
"parameters": [
	{
		"name": "first",
		"in": "query",
		"description": "...",
		"required": true,
		"schema": {
			"type": "string"
		}
	},
	{
		"name": "second",
		"in": "query",
		"description": "...",
		"required": true,
		"schema": {
			"type": "string"
		}
	}
]
```

springdoc:
Hi, 

 @ModelAttribute is related to web views. 
 Its not supposed to be used on REST APIs.
 We are not planning to support it.

Mattias-Sehlstedt:
> > Hi,
> > [@ModelAttribute](https://github.com/ModelAttribute) is related to web views. Its not supposed to be used on REST APIs. We are not planning to support it.
> 
> I know this issue is closed, but this means that it's not possible to use `@ModelAttributes` with `springdoc-openapi`. I understand that one might not be "supposed" to use ModelAttributes in REST APIs, but one _CAN_, and it's powerful (unless there is another way to achieve the same effect by other means).
> 
> So from what I understand, one would need to choose one or the other... is there no workaround that can be done that enables us to be able to use both?

@roberto-welzel, are you still looking for an answer to this or have you found a solution?

alkismavridis:
Hi there, can you provide some links or explanation why it is not supposed to be used in rest apis and why?
[...]

PR #505, opened 2020-03-23
Extract parameters from parameter object

### LINKED ISSUE #120: Add support for object as request parameter container
Spring supports that an object is used to define query parameters as shown below:

**Controller**
```
@RestController
public class MealPartiesAdminController {
    @GetMapping(value = "/parties/{id}")
    public MealParty getMealParty(@PathVariable("id") long mealPartyId, MealPartiesQuery query){
        return this.mealPartyAdminService.getMealParty(mealPartyId, query);
    }
}
```

**Model**
```
public class MealPartiesQuery {
    private String name;
    private int membersCount;
}
```

Both `name` and `membersCount` are now endpoint query parameters. However, the generated swagger information only shows the query as an object. I've also added `@ModelAttribute` to `query` parameter in the controller but it still shows it as an object.
Am I missing something or there is no support for this currently?

zarebski-m:
Fixes #120, #268, #162, #119

This PR introduces a new Springdoc annotation `org.springdoc.api.annotations.ParameterObject`. Request parameter annotated with it are not added as a request parameter themselves, but rather each field of the parameter is added as a separate request parameter. This is compatible with Spring MVC request parameters mapping to POJO object.

See test case 102 in springdoc-openapi-webmvc-core.

Consider the following classes:

```java
class QueryParams {
    private String name;
    private String value;
}

@RestController
class TestController {
    @GetMapping("pojo-params")
    public void testMethod(@ParameterObject QueryParams queryParams) {}
}
```

In generated docs, the `/pojo-params` endpoint defines two request parameters: `name` and `value`, instead of single `queryParams` one.

This also allows deprecation of the workaround for `Pageable` with `PageableAsQueryParam` annotation – we can simply use `@ParameterObject Pageable pageable` on method parameter.

See test case 7 on springdoc-openapi-data-rest.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+tab_width = 4
+indent_style = tab
+
+[*.json]
+indent_style = space
+indent_size = 2
--- a/springdoc-openapi-common/src/main/java/org/springdoc/api/annotations/ParameterObject.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/api/annotations/ParameterObject.java
@@ -1,0 +1,10 @@
+package org.springdoc.api.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.PARAMETER)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface ParameterObject {}
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/AbstractRequestBuilder.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/AbstractRequestBuilder.java
@@ -18,6 +18,29 @@
 
 package org.springdoc.core;
 
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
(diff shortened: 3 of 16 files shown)


zarebski-m:
Thank you, I've applied https://github.com/spring-projects/spring-boot/blob/master/idea/codeStyleConfig.xml to my IDE and reformatted my code.

bnasslahsen:
Its now merged with master.

prondzyn:
@bnasslahsen when are you planning to release this feature?

bnasslahsen:
@prondzyn , 

Should be available, no later than next Monday, 06th April.

destebanm:
HI!!

Thanks for the new annotation @ParameterObject!

Nevertheless, we are facing an issue when using it. Every attribute of the POJO is being set as required. Is there any way to avoid that?
image
image

image


Thanks!

bnasslahsen:
@destebanm, 

 This is already fixed. https://github.com/springdoc/springdoc-openapi/issues/541.
 You can just use v1.3.2.

destebanm:
WOW, it is working fine with 1.3.2!

Thanks!

dmittriy13:
Hi, thanks for your work, I ran into a problem: inherited fields are not taken into account.
image

[Here](https://stackoverflow.com/a/1042827) is an example of how to handle this case.
I would be very grateful if you correct this in the next version.

Ichigo85:
+1
I also need inheritance working for method parameters.

bnasslahsen:
@Ichigo85, 

 Its already released. You can use v1.3.6.

svetlana-nikitina:
@zarebski-m is it possible to have the same behaviour for POST requests, as described here: https://github.com/springdoc/springdoc-openapi/issues/693

PR #532, opened 2020-04-02
Fixing encoding section in multipart request Fixes #531

### LINKED ISSUE #531: Encoding section for multipart requests do not work
**Describe the bug**
swagger.io documentation has support for adding your own `Content-Type` for specific parts of `multipart / form-data` request and `springdoc` also has this support, but when it's set, it doesn't appear in the result JSON

 - https://swagger.io/docs/specification/describing-request-body/multipart-requests/#content-type

**To Reproduce**
```java
@RestController
public class HelloController {

    @PostMapping(value = "/test/103", consumes = MediaType.MULTIPART_FORM_DATA_VALUE)
    @Operation(
            requestBody = @RequestBody(content = @Content(encoding = @Encoding(name = "body", contentType = "application/json")))
    )
    public String postMyRequestBody(
            @RequestPart("body") ExampleBody body,
            @RequestParam("file") MultipartFile file
    ) { return null; }
}
```

**Expected behavior**
In the JSON response, we should see additional section
```
"encoding":{ "body":{ "contentType":"application/json" } }
```

patrykpacewicz:


patrykpacewicz:
Super happy that I could help. Thanks for so fast response. Do you know when next version (and this change) will be released ?

bnasslahsen:
should be available, no later than next Monday, 06th April.

PR #593, opened 2020-04-21
issue-588-partialfix1

raghuraman1:
@bnasslahsen
This will prevent non generic controller advices from being applied
generically. 
Next step: After this will work on non generic controller advices targetting specific controllers

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java
@@ -134,9 +134,15 @@ public abstract class AbstractOpenApiResource extends SpecFilter {
 					.collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (a1, a2) -> a1));
 
 			Map<String, Object> findControllerAdvice = openAPIBuilder.getControllerAdviceMap();
+			Map<Boolean, List<Map.Entry<String, Object>>> partitionedControllerAdviceEntries = 
+					findControllerAdvice.entrySet().stream().collect(Collectors.partitioningBy(
+							entry -> OpenAPIBuilder.isTargettingControllerAdvice(entry)));
+			Map<String, Object> targettingControllerAdvice = controllerAdviceEntryListToMap(partitionedControllerAdviceEntries, true);
+			//targettingControllerAdvice must be used in getPaths later maybe after some transformng logic
+			Map<String, Object> genericControllerAdvice = controllerAdviceEntryListToMap(partitionedControllerAdviceEntries, false);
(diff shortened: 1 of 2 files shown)


raghuraman1:
@bnasslahsen Just thought will add some notes:
1. Created OpenAPIBuilder.isTargettingControllerAdvice mainly to avoid bringing in ControllerAdvice into AbstractOpenApiResource.
2.  In OpenAPIBuilder.isTargettingControllerAdvice have a look at this comment //not expecting this else and //ignoring will check for suggestion.
I should think its safe to ignore the else. But what will you suggest?

raghuraman1:
@bnasslahsen I did a manual test using the ui. will repeat using json.

raghuraman1:
behaviour is as expected. The ControllerAdvice am using is like this:
@ControllerAdvice(assignableTypes = PersonController.class)
The jsons are as expected.
Attaching them
[using1.3.4-withoutfix.json.txt](https://github.com/springdoc/springdoc-openapi/files/4516915/using1.3.4-withoutfix.json.txt)

[withfix.json.txt](https://github.com/springdoc/springdoc-openapi/files/4516920/withfix.json.txt)

bnasslahsen:
This looks OK. If you could just update your PR, with these elements by adding the test on the springdoc-openapi-webmvc-core module.

If its not possible you can share all the classes that help to reproduce.

raghuraman1:
It was a manual test. I will give you the code I have. You will have to change the pom.xml yourself. For 1.3.5-SNAPSHOT you will have clone from my fork I suppose- https://github.com/raghuraman1/springdoc-openapi (you would know best)
You will have to change the @ControllerAdvice() annotation manually to @ControllerAdvice(assignableTypes = PersonController.class)
I am attaching the sample  project source code
[sample-sourcecode.zip](https://github.com/springdoc/springdoc-openapi/files/4517102/sample-sourcecode.zip)

Hope that helps.

R

bnasslahsen:
@raghuraman1,

 While using @ControllerAdvice(assignableTypes = PersonController.class), the advice should be applied to PersonController which is not the case in the file withfix.json.txt.

I will have a look to improve the fix.

raghuraman1:
I have given u a partial fix. What you mentioned is the next step.
I wanted the easy parts done first.

raghuraman1:
@bnasslahsen  Just making sure- In part II of the fix we must take care of assignableTypes, basePackageClasses, basePackages and not just assignableTypes.

bnasslahsen:
@raghuraman1, 

ok. it's taken care. The fix is now available on master.

raghuraman1:
Thanks.

raghuraman1:
@bnasslahsen I tried using master and my sample code. Could not see the fix. Tried testing also

Issue #590, opened 2020-04-21
@ParameterObject  support nested parameter objects


luhuiguo:
When use Jhipster generator entity filtering
[https://www.jhipster.tech/entities-filtering/](https://www.jhipster.tech/entities-filtering/)

Parameter object 
```java
public class BookCriteria implements Serializable, Criteria {
  private LongFilter id;
  private StringFilter name;
  ......
```
Inner object
```java
public class StringFilter extends Filter<String> {
    private String contains;
    private String doesNotContain;
    ......
```

```java
  @GetMapping("/books")
  public ResponseEntity<List<BookDTO>> getBooks(@ParameterObject BookCriteria criteria, @ParameterObject Pageable pageable) {
    log.debug("REST request to get Book by criteria: {}", criteria);
    Page<BookDTO> page = bookQueryService.findByCriteria(criteria, pageable);
    ......
```

** Springfox 
For each entity, you can enable filtering in the entity generator, and after, you can call your /api/my-entity GET endpoint with the following parameters :

For each xyz field
xyz.equals=someValue
To list all the entities, where xyz equals to ‘someValue’
xyz.in=someValue,otherValue
To list all the entities, where xyz equals to ‘someValue’ or ‘otherValue’
xyz.specified=true
To list all the entities, where xyz is not null, specified.
xyz.specified=false
To list all the entities, where xyz is null, unspecified.
If xyz’s type is string:
xyz.contains=something
To list all the entities, where xyz contains ‘something’.
......

** Springdoc @ParameterObject does't support nested parameter objects

```json
id : {
  "equals": 0,
  "notEquals": 0,
  "specified": true,
  "in": [
    0
  ],
  "greaterThan": 0,
  "lessThan": 0,
  "greaterThanOrEqual": 0,
  "lessThanOrEqual": 0,
  "greaterOrEqualThan": {
    "equals": 0,
    "notEquals": 0,
    "specified": true,
    "in": [
      0
    ],
    "greaterThan": 0,
    "lessThan": 0,
    "greaterThanOrEqual": 0,
    "lessThanOrEqual": 0
  },
  "lessOrEqualThan": {
    "equals": 0,
    "notEquals": 0,
    "specified": true,
    "in": [
      0
    ],
    "greaterThan": 0,
    "lessThan": 0,
    "greaterThanOrEqual": 0,
    "lessThanOrEqual": 0
  }
}
```

bnasslahsen:
@luhuiguo, 

[...]

PR #596, opened 2020-04-22
Get fields of superclass for parameter objects

zarebski-m:


bnasslahsen:
Thank you @zarebski-m for your enhancement.
Would you be able to add support for nested parameter objects as requested in #590?

zarebski-m:
I could give it a try, but not before next month.

Issue #610, opened 2020-04-25
ApiResponse DEFAULT_DESCRIPTION Improvement


raghuraman1:
Hi @bnasslahsen

Remember below code in ControllerAdvice

![image](https://user-images.githubusercontent.com/14346578/80275501-3d36b780-86ff-11ea-972b-923c08185dc9.png)

   
Look at the corresponding responses:

![image](https://user-images.githubusercontent.com/14346578/80275355-4b380880-86fe-11ea-88f6-17cd9a75ad62.png)

Expectation is that instead of "defaultResponse" it should pickup the message from HttpStatus.getReasonPhrase().

Raghu

raghuraman1:
@bnasslahsen The Travis test failed for related PR https://github.com/springdoc/springdoc-openapi/pull/611 and I presume thats a good thing. The related tests will need changing.

bnasslahsen:
@raghuraman1, 

 Ok, i will see the impacts on the tests. 
 The improvement you propose, is not only related to the controllerAdvice.

raghuraman1:
Agreed. I was just writing that its beyond ControlerAdvice.

raghuraman1:
@bnasslahsen  I can fix the tests as part of the PR

bnasslahsen:
@raghuraman1, 

 It's ok. I have already started the merge.

raghuraman1:
I just noticed what you typed. Was committing the test changes

raghuraman1:
@bnasslahsen  I do wish I could have some PR committed in my name also. :)

bnasslahsen:
@raghuraman1, 

 Next time 👍

raghuraman1:
@bnasslahsen This time it was like a robbery from my point of view. :)
But you guys are making a great product.
I

PR #611, opened 2020-04-25
Fix for https://github.com/springdoc/springdoc-openapi/issues/610

raghuraman1:
Fix for https://github.com/springdoc/springdoc-openapi/issues/610


bnasslahsen:
@raghuraman1 are now merged with master.

PR #628, opened 2020-04-29
Make use of @Deprecated annotations for parameters and model fields

zarebski-m:
Currently `@Deprecated` annotation affects only whole API operations.

Within this PR I'm introducing a mechanism to mark model attributes and operation parameters as deprecated as well, using the standard `@Deprecated` annotation.


bnasslahsen:
Its now merged with master.

PR #648, opened 2020-05-05
Register model converters only if they are not registered already - fix #647

### LINKED ISSUE #647: ModelConverters are registered multiple times
Hi,

We're using springdoc-openapi 1.3.4 and wrote some tests for swagger doc generation. As all our tests are annotated with @SpringBootTest, the model converters will be registered for every test class. In order to mitigate this, we would need to use your approach from test.org.springdoc.api.AbstractSpringDocTest and unregister all model converters on @AfterAll.

However, this approach is not applicable for us, as we run our tests in parallel.

There would be useful if the model converters are registered only if they are not already.  This way that cleanup on @AfterAll would not be necessary anymore for sequential tests and for the parallel tests would be covered out of the box.

valentinabojan:


bnasslahsen:
Thank you @valentinabojan for your contribution !

PR #753, opened 2020-06-29
Openapi Server Object url supports reverse proxy

### LINKED ISSUE #752: Openapi Server Object url supports reverse proxy
There is a webflux-functional endpoint A service. Proxy via Spring Cloud Gateway.
A service is behind the gateway, so use the following solution
  https://springdoc.org/faq.html#how-can-i-deploy-springdoc-openapi-ui-behind-a-reverse-proxy

After setting the `forward-headers-strategy: framework` property
The webflux-functional endpoint request starts at 404.
 Specific information can be viewed spring-projects/spring-framework#25270

oursy:
* resolves #752

### DIFF
--- a/springdoc-openapi-webmvc-core/src/test/resources/results/app99.json
+++ b/springdoc-openapi-webmvc-core/src/test/resources/results/app99.json
@@ -6,7 +6,7 @@
   },
   "servers": [
     {
-      "url": "http://localhost",
+      "url": "http://localhost:80",
       "description": "Generated server url"
     }
   ],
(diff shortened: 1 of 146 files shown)


springdoc-openapi-webmvc-core/src/test/resources/results/app99.json
```diff
@@ -6,7 +6,7 @@
   },
   "servers": [
     {
-      "url": "http://localhost",
+      "url": "http://localhost:80",
```
maciejkopecpl, line 9: Why in all test results have you added port 80?

oursy:
This problem will be addressed in Spring Framework, see spring-projects/spring-framework#25270

bnasslahsen:
Grreat @oursy.

Thank you for sharing the feedback of Spring framework team.

PR #755, opened 2020-06-29
Marker interface for swagger ui index transformer

### LINKED ISSUE #745: change SwaggerIndexTransformer and SwaggerConfig design to ease customization
When people want to extend the behavior of Swagger UI transformers (``SwaggerIndexTransformer`` declared in `org.springdoc.webmvc.ui.SwaggerConfig` and `org.springdoc.webflux.ui.SwaggerConfig` they have to inherit from your specialized classes instead of just using the provided abstract support class (``AbstractSwaggerIndexTransformer``).

I think that the design could be improved and the **signatures used in `SwaggerConfig` classes modified to ease work for reuse**.

In my case I want to inject css file and js file (used for OIDC) to the document. I created a support class similar to:

```java
public class MySwaggerIndexTransformerSupport extends AbstractSwaggerIndexTransformer {
// other stuff

	public Resource transform(Resource resource, ResourceTransformerChain resourceTransformerChain) {
// I mask the IOException management
// use methods from AbstractSwaggerIndexTransformer 
// execute my own transformation
// return resource
	}
}
```
And to use it I create a Bean:
```java
/*
 * Keep the parent only to deactivate autoconfiguration from springdoc
 */
public class MySwaggerIndexTransformer extends SwaggerIndexTransformer implements ResourceTransformer {

	private MySwaggerIndexTransformerSupport support;

    public MySwaggerIndexTransformer(SwaggerUiConfigProperties swaggerUiConfig, SwaggerUiOAuthProperties swaggerUiOAuthProperties, ObjectMapper objectMapper) {
        super(swaggerUiConfig, swaggerUiOAuthProperties, objectMapper);
		support = new MySwaggerIndexTransformerSupport(swaggerUiConfig, swaggerUiOAuthProperties, objectMapper);
    }

    @Override
    public Mono<Resource> transform(ServerWebExchange serverWebExchange, Resource resource, ResourceTransformerChain resourceTransformerChain) {
        return Mono.just(support.transform(resource, resourceTransformerChain));
    }

}
```

I have 2 proposals based on this use case.

1. introduce an interface `SwaggerIndexPageTransformer` declared in common module and used in `SwaggerConfig` classes 
```java
	@Bean
	@ConditionalOnMissingBean
	SwaggerIndexPageTransformer ...
```
The inconvenience is that it cannot extend `ResourceTransformer` since `ResourceTransformer` exist for *servlet* and *reactive* resources. Since transformation only uses the Resource you could declare a reduced signature:
`Resource transform(Resource resource);`

2. change ``AbstractSwaggerIndexTransformer`` to a support class used by composition instead of inheritance to simplify reuse

With this, the specialized behavior would not have to inherit the current `SwaggerIndexTransformer` but 'only' need to implement an interface.

- Instead of introducing a new interface (that would break the current contract of the `SwaggerConfig` classes you could use a named bean: 
```java
@Bean
@ConditionalOnMissingBean(name="swaggerUiIndexPageTransformer")
ResourceTransformer indexPageTransformer(...) {...}
```
I think this is less readable than the proposed evolution and it also breaks the current contract.

The other possibility I see is to declare the `AbstractSwaggerIndexTransformer` class in signature (instead of the specialized implementation 

**Additional context**
I'm willing to work on this if you agree on these changes.

natrem:
close #745

as explained [here](https://github.com/springdoc/springdoc-openapi/issues/745#issuecomment-651041778) it eases providing a custom SwaggerIndexTransformer by introducing an intermediary interface.

@bnasslahsen please note the 2nd commit. I factorized some behavior in the abstract class. The rationale is that if you have to duplicate this in mvc and reactive implementations then application creating a new transformation (for instance the [linked example](https://github.com/springdoc/springdoc-openapi/issues/745#issuecomment-651041778)) will face the same issue and may have to rewrite same code.
If you're not comfortable with it I don't mind dropping this part; since you have `overwriteSwaggerDefaultUrl` and  `addInitOauth` I supposed we can add other methods related to the default implementation.

Since you rely on integration tests I did not add specific unit tests on the autoconfiguration but I can if you wish (I've done a few lately so it won't take long)

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/ui/AbstractSwaggerIndexTransformer.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/ui/AbstractSwaggerIndexTransformer.java
@@ -31,6 +31,8 @@ import org.apache.commons.lang3.StringUtils;
 import org.springdoc.core.Constants;
 import org.springdoc.core.SwaggerUiConfigProperties;
 import org.springdoc.core.SwaggerUiOAuthProperties;
+import org.springframework.core.io.Resource;
+import org.springframework.util.CollectionUtils;
 
 /**
  * The type Abstract swagger index transformer.
@@ -108,5 +110,22 @@ public class AbstractSwaggerIndexTransformer {
 	 */
 	protected String overwriteSwaggerDefaultUrl(String html) {
 		return html.replace(Constants.SWAGGER_UI_DEFAULT_URL, StringUtils.EMPTY);
+	}
+
+	protected boolean hasDefaultTransformations() {
+		boolean oauth2Configured = !CollectionUtils.isEmpty(swaggerUiOAuthProperties.getConfigParameters());
+		return oauth2Configured || swaggerUiConfig.isDisableSwaggerDefaultUrl();
+	}
+
+	protected String defaultTransformations(InputStream inputStream) throws IOException {
+		String html = readFullyAsString(inputStream);
+		if (!CollectionUtils.isEmpty(swaggerUiOAuthProperties.getConfigParameters())) {
+			html = addInitOauth(html);
+		}
+		if (swaggerUiConfig.isDisableSwaggerDefaultUrl()) {
+			html = overwriteSwaggerDefaultUrl(html);
+		}
+
+		return html;
 	}
 }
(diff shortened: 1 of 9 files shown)


bnasslahsen:
Thank you @natrem for your contribution 👍 
I couldn't resist to renaming the classes SwaggerIndexPageTransformer and calling the interface SwaggerIndexTransformer. which makes code much cleaner and less impacted files.

It will be available for the next release v1.4.3.

natrem:
I agree the renaming is better but I didn't dare.
thanks again for your job in maintaining these useful libraries

PR #802, opened 2020-07-27
fix comment on api-docs.enabled in README.MD

kristofpeto-tesco:


bnasslahsen:
Thank you @kristofpeto-tesco for your contribution.

PR #815, opened 2020-08-04
Allow swagger-ui property filter to be an empty string

gjorgievskivlatko:
At the moment if we set the swagger-ui property filter to `true` then it is not working as expected.
I already reported this to the swagger-ui team at:
https://github.com/swagger-api/swagger-ui/issues/6276
We should be able to set the filter property to empty string because in that case the filter is working properly.


bnasslahsen:
Thank you @gjorgievskivlatko for your contribution.
It will be integrated on the next release 1.4.4.

gjorgievskivlatko:
Thank you @bnasslahsen.

PR #854, opened 2020-09-04
Add pre-loading setting to load OpenAPI early

### LINKED ISSUE #850: Allow for pre-loading OpenAPI
In our project generating Swagger for large instances can take a long time, causing a long delay for the user (30+ seconds). After upgrading from `1.4.4` to SpringDoc `1.4.6` our existing solution no longer works. We currently have a custom pre-loader that forces OpenAPI to generate swagger by making a HTTP request to SpringDoc when the application starts

Our proposal would be to allow for pre-loading swagger when starting the application, using a method call.

Dodecahedra:
Add configuration setting to `SpringDocConfigProperties` to enable pre-loading of OpenAPI. This setting is false by default. When instantiating `AbstractOpenApiResource` the OpenAPI is generated in a separate thread. Fixes #850


bnasslahsen:
Thank you @Dodecahedra for your contribution. 
It's now merged with master.

PR #856, opened 2020-09-04
Prevent API component schema description from being overwritten Fixes #852

### LINKED ISSUE #852: API component schema description incorrectly overwritten by API parameter description
**Describe the bug**
The description for a shared API component schema is overwritten by the description for an API parameter.

**To Reproduce**
Create a class annotated with the @Schema annotation and specify a description. Create a method with a parameter with the type of this class annotated with the @Parameter annotation with a description. The description for the shared class is overwritten with description for the parameter.

**Organization.java**
```
package digital.inception.party;

import com.fasterxml.jackson.annotation.JsonIgnore;
import com.fasterxml.jackson.annotation.JsonInclude;
import com.fasterxml.jackson.annotation.JsonProperty;
import com.fasterxml.jackson.annotation.JsonPropertyOrder;
import io.swagger.v3.oas.annotations.media.Schema;
import java.util.UUID;

@Schema(
    description =
        "This is the description being overwritten")
@JsonInclude(JsonInclude.Include.NON_NULL)
@JsonPropertyOrder({"id", "name"})
public class Organization {

  @Schema(
      description =
          "The Universally Unique Identifier (UUID) uniquely identifying the organization",
      required = true)
  @JsonProperty(required = true)
  private UUID id;

  @Schema(description = "The name of the organization", required = true)
  @JsonProperty(required = true)
  private String name;

  public Organization() {}

  public UUID getId() {
    return id;
  }

  public String getName() {
    return name;
  }

  public void setId(UUID id) {
    this.id = id;
  }

  public void setName(String name) {
    this.name = name;
  }
}
```

**API Method**

```
  @Operation(summary = "Create the organization", description = "Create the organization")
  @ApiResponses(
      value = {
        @ApiResponse(
            responseCode = "204",
            description = "The organization was created successfully"),
        @ApiResponse(
            responseCode = "400",
            description = "Invalid argument",
            content =
                @Content(
                    mediaType = "application/json",
                    schema = @Schema(implementation = RestControllerError.class))),
        @ApiResponse(
            responseCode = "409",
            description = "An organization with the specified ID already exists",
            content =
                @Content(
                    mediaType = "application/json",
                    schema = @Schema(implementation = RestControllerError.class))),
        @ApiResponse(
            responseCode = "500",
            description =
                "An error has occurred and the request could not be processed at this time",
            content =
                @Content(
                    mediaType = "application/json",
                    schema = @Schema(implementation = RestControllerError.class)))
      })
  @RequestMapping(
      value = "/organizations",
      method = RequestMethod.POST,
      produces = "application/json")
  @ResponseStatus(HttpStatus.NO_CONTENT)
  public void createOrganization(
      @Parameter(name = "organization", description = "This description will overwrite the existing description", required = true)
          @RequestBody
          Organization organization)
      throws InvalidArgumentException, DuplicateOrganizationException, PartyServiceException {

   ...

  }
```

I believe the problem is in the **calculateRequestBodySchema** method of the **org.springdoc.core.GenericParameterBuilder** class.

The code below retrieves the existing component schema and replaces the description with the description for the API parameter.

```
if (schemaN != null && StringUtils.isEmpty(schemaN.getDescription()) && parameterInfo.getParameterModel() != null) {
			String description = parameterInfo.getParameterModel().getDescription();
			if (schemaN.get$ref() != null && schemaN.get$ref().contains(AnnotationsUtils.COMPONENTS_REF)) {
				String key = schemaN.get$ref().substring(21);
				Schema existingSchema = components.getSchemas().get(key);
				existingSchema.setDescription(description);
			}
			else
				schemaN.setDescription(description);
		}
```

I am not sure if there is a valid reason for doing this, perhaps if the description doesn't exist for the shared component schema this might be a valid approach. If a description does exist for shared component schema then this behaviour does not seem correct. I am happy to submit a pull request that changes this behaviour to only change the description if it has not already been set for the shared component schema.

I am using springdoc-openapi-common 1.4.6.

marcusportmann:
This fixes #852 by only overwriting the shared schema description if one is not already present.

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/GenericParameterBuilder.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/GenericParameterBuilder.java
@@ -304,7 +304,10 @@ public class GenericParameterBuilder {
 			if (schemaN.get$ref() != null && schemaN.get$ref().contains(AnnotationsUtils.COMPONENTS_REF)) {
 				String key = schemaN.get$ref().substring(21);
 				Schema existingSchema = components.getSchemas().get(key);
-				existingSchema.setDescription(description);
+
+				if (StringUtils.isEmpty(existingSchema.getDescription())) {
+					existingSchema.setDescription(description);
+				}
 			}
 			else
 				schemaN.setDescription(description);


bnasslahsen:
@marcusportmann, 

 Thank you for your PR. 
 It has just been changed a little here https://github.com/springdoc/springdoc-openapi/commit/bdfa6eaefafee626e6b80383dc47bd1e665e722d as explained here https://github.com/springdoc/springdoc-openapi/issues/852#issuecomment-687768330

Issue #885, opened 2020-09-30
Flaky operationIds for controller-methods with same name


kaibra:
**Describe the bug**
Given a controller with methods of the same name, the operationIds are automatically build to be unqiue by appending an increasing number (eg "_1") to the method-name. The ids are not being build deterministically. Two runs can lead to different operationIds for a path in the spec.  

**To Reproduce**
```Java
@RestController
public class OperationIdController {

    @GetMapping(path = "/test_0")
    public String opIdTest() {
        return "";
    }

    @GetMapping(path = "/test_1")
    public String opIdTest(@RequestParam String param) {
        return "";
    }

    @GetMapping(path = "/test_2") 
    public String opIdTest(@RequestParam Integer param) {
        return "";
    }

    @GetMapping(path = "/test_3")
    public String opIdTest(@RequestParam Boolean param) {
        return "";
    }

}
```
If you have a look at the OpenAPI calculated for this Controller, operationIds are flaky and can change for the given paths.

**Expected behavior**
Even though it may not make sense to have such a controller set up, the behaviour should be deterministic.
A given Controller should always lead to the same OpenAPI being calculated. 

**Additional context**
There is already a sorting in place here: https://github.com/springdoc/springdoc-openapi/blob/master/springdoc-openapi-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java#L413 before `calculatePath` is called.
But not here: https://github.com/springdoc/springdoc-openapi/blob/master/springdoc-openapi-webmvc-core/src/main/java/org/springdoc/webmvc/api/OpenApiResource.java#L244
And here: https://github.com/springdoc/springdoc-openapi/blob/master/springdoc-openapi-webflux-core/src/main/java/org/springdoc/webflux/api/OpenApiResource.java#L193

The EntrySet/Map from RequestMappingHandlerMapping does not seem to have a deterministic order.

PR #886, opened 2020-09-30
fix for flaky operationIds

kaibra:
This PR fixes https://github.com/springdoc/springdoc-openapi/issues/885

### DIFF
--- a/springdoc-openapi-webmvc-core/src/main/java/org/springdoc/webmvc/api/OpenApiResource.java
+++ b/springdoc-openapi-webmvc-core/src/main/java/org/springdoc/webmvc/api/OpenApiResource.java
@@ -20,12 +20,7 @@
 
 package org.springdoc.webmvc.api;
 
-import java.util.HashSet;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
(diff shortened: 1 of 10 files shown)


springdoc-openapi-webmvc-core/src/main/java/org/springdoc/webmvc/api/OpenApiResource.java
```diff
@@ -265,6 +262,12 @@ protected void calculatePath(Map<String, Object> restControllers, Map<RequestMap
 		}
 	}
 
+	private Comparator<Map.Entry<RequestMappingInfo, HandlerMethod>> byReversedRequestMappingInfos() {
```
kaibra, line 265: Wasn't sure, if you want to share this somehow between webflux and webmvc?!

bnasslahsen: It would have been an ideal to refactor it in the AbstractOpenApiResource. 
But, it doesn't seem to be simple to achieve.
Your solution looks fine!

bnasslahsen (approved):
Thank you @kaibra for your excellent contribution.

PR #894, opened 2020-10-07
Ignoring annotation should support meta-annotations

michaldo:
If @A is ignored and @B is meta-annotation containing @A, then @B should be also ignored


bnasslahsen:
> If @A is ignored and @b is meta-annotation containing @A, then @b should be also ignored

Thank you @michaldo for your contribution.
Could you please complete your conctribution, by making sure that all the tests are passing?

PR #898, opened 2020-10-12
Spring Security support out of the box

michaldo:
It would be easier if Springdoc does not need extra dependencies to support standard Spring Security conventions

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/AbstractRequestBuilder.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/AbstractRequestBuilder.java
@@ -130,6 +130,11 @@ public abstract class AbstractRequestBuilder {
 		PARAM_TYPES_TO_IGNORE.add(SessionStatus.class);
 		PARAM_TYPES_TO_IGNORE.add(UriComponentsBuilder.class);
 		PARAM_TYPES_TO_IGNORE.add(RequestAttribute.class);
+		try {
+			PARAM_TYPES_TO_IGNORE.add(Class.forName("org.springframework.security.core.Authentication"));
+		} catch (ClassNotFoundException e) {
+			// Spring security not present
+		}
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


bnasslahsen:
@michaldo,

 Thank you for you proposal of refactoring.
 There is missing functionalities on your PR, like the support for displaying the login-endpoint provided by spring-security.
 Additionally, projects that relies on OAuth2 client, will still have to add the dependency.

 For clarity, it better to rely on the current configuration, as it makes responsibilities of each module much clear.

michaldo:
In my company I maintain springdoc for many REST-based application. Some of them use Authentication, some has open access.
I resolved client identification by  `OpenAPI#addSecurityItem` - https://springdoc.org/faq.html#how-do-i-add-authorization-header-in-requests
I have configuration switch to enable/disable `OpenAPI#addSecurityItem`. That way single component without extra dependencies can be reused by various applications.

If I add `springdoc-openapi-security` to my component, I will add `spring-security-core` to all applications - I would like avoid adding unnecessary dependencies.

Please remember that there are plenty Authentication methods, for example JWT, Kerberos-LDAP etc. `springdoc-openapi-security` will never handle all cases.

## Summary
`springdoc-openapi` primary goal is transform Spring REST controller methods into Swagger UI. Method parameter `Authentication` etc. is never REST parameter and should be ignored. That is my natural expectation.
`OpenAPI#addSecurityItem` is good tool to resolve client identification.

I appreciate I can use  `springdoc-openapi-security` with useful Spring login endpoint
I will appreciate I could NOT use `springdoc-openapi-security` and have natural expectation fulfilled

bnasslahsen:
@michaldo, 

 Each module has a role. If an application doesn't use spring-security, you don't add springdoc-openapi-security.

 For your last point, if you don't want rely on springdoc-openapi-security, because your only requirement, is to ignore some simple spring-security annotations, you can directly use:

-  `SpringDocUtils.addAnnotationsToIgnore`
-  `AbstractRequestbuilder.addRequestWrapperToIgnore`
-  `ConverterUtils.addResponseWrapperToIgnore`

Note, that the code is Open, and you can extend or adapt any configuration.

michaldo:
Indeed, I can use `SpringDocUtils.addAnnotationsToIgnore` and ignore `Authentication` without `springdoc-openapi-security` dependency.

I created this PR because I think everybody want ignore  `Authentication` and the ignoring should work out of the box.

PR #917, opened 2020-10-31
fix issue 907. Hidden controller showing up in swagger UI when spring…

hismaili:
…doc.show-actuator is enabled


hismaili:
checks failed

PR #961, opened 2020-12-02
Include an automatic module name in generated manifests

kashike:
http://blog.joda.org/2017/05/java-se-9-jpms-automatic-modules.html


bnasslahsen:
@kashike, 

 Great contribution 👍

PR #978, opened 2020-12-09
Typesave serialization of openapi spec

kevinraddatz:
I was trying to integrate spring cloud config into our project and used the openAPI() bean for testing.
```java
@Bean
@RefreshScope
public OpenAPI openAPI() {
    return new OpenAPI()
        .info(new Info()
                    .title("title)
                    .description("description")
                    .version("1.0.0")
         );
}
```

By default spring uses the proxy method `TARGET_CLASS` to generate a refreshScope bean which adds multiple properties which will be serialized by the objectMapper. Serialization of those properties results in an infinite loop and thus in a StackOverflowError.

With these changes the openAPI object will be serialized for the OpenAPI class, ignoring the properties added by spring.


bnasslahsen (approved):
thank you @kevinraddatz for your contribution;

PR #987, opened 2020-12-11
Add explode support in combination with arrayShema annotation

### LINKED ISSUE #984: GenericParameterService#isExplodable method should recognise @ArraySchema as well
**Describe the bug**
It seems that sprindoc is not allowing the usage of explode attribute on Parameter, when it s combined with an ArraySchema annotation.

I ve been debugging the issue and spotted that GenericParameterService has a method isExplodable that only supports @Schema annotation, which is not enough in case of Array, because you can't define the items type in the Schema annotation.

**To Reproduce**
- What version of spring-boot you are using? 2.4.0
- What modules and versions of springdoc-openapi are you using? springdoc-openapi-ui 1.5.0

``` 
@GetMapping(value = "helloWorld")
@Operation(summary ="helloWorld", description = "helloWorld")
public List<String> helloWorld(
    @Parameter(description = "Dummy list", required = true, explode = Explode.FALSE, , array = @ArraySchema( schema = @Schema(type = "string")), style = ParameterStyle.FORM) @RequestParam(dummyList) List<String> dummyList) { 
   return dummyList; 
} 
```

Produce parameter like

```
          {
            "name": "dummyList",
            "in": "query",
            "description": "Dummy list",
            "required": true,
            "style": "form",
            "schema": {
               "type": "array",
               "items": {
                   "type": "string"
               }
            }
          }
```

**Expected behavior**
So basically it is expected that @ArraySchema should be treated as @Schema type array, to ensure that array in query parameters are correctly typed in the open api spec.

So it should produce a parameter like : 

`Should produce a parameter like : 

```
          {
            "name": "dummyList",
            "in": "query",
            "description": "Dummy list",
            "required": true,
            "style": "form",
            "explode": false,
            "schema": {
               "type": "array",
               "items": {
                   "type": "string"
               }
            }
          }
```

coutin:
Fix #984 

This add supports to explode open api attribute for array query parameters using the arraySchema annotation.

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/GenericParameterService.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/GenericParameterService.java
@@ -54,9 +54,11 @@ import org.springframework.core.ResolvableType;
 import org.springframework.core.io.Resource;
 import org.springframework.web.multipart.MultipartFile;
 
+import static java.util.Objects.isNull;
+
 /**
  * The type Generic parameter builder.
- * @author bnasslahsen
+ * @author bnasslahsen, coutin
  */
 @SuppressWarnings("rawtypes")
 public class GenericParameterService {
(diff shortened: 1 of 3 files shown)


bnasslahsen:
@coutin, 

 Thank you for your contribution;
 I have made some change, as the condition  `isNull(arraySchema)` is in fact always false.

PR #992, opened 2020-12-14
OpenApiCustomizer

seregamorph:
The problem: `OpenApiCustomiser` has typo in both interface name and method. The trouble: back compatibility (BC) contracts.
Proposed solution: `OpenApiCustomiser` (has `customise()` method) extracted to super-interface `OpenApiCustomizer` (has `customize()` method). Deprecated `OpenApiCustomiser` has `customize()` method implementation as delegate to obsolete `customise()`, saving BC.

Kept binary compatibility:
AbstractOpenApiResource constructor (same as its subclasses) now accepts `Optional<List<OpenApiCustomizer>>` instead of `Optional<List<OpenApiCustomiser>>`, so in case of iterating the runtime is not broken (cast is possible). It may cause compile issues (`Optional<List<OpenApiCustomiser>>` is not assignable), but:
- this way we emphasize the problem to a developer that uses this API
- binary compatibility kep

:warning: Compromising binary incompatibility: :warning:
ActuatorOpenApiCustomizer: now does not extend `OpenApiCustomiser`, only `OpenApiCustomizer`. (may be broken in case if this class has subclasses, but it's assumed that it's not reasonable)

:exclamation: Broken binary incompatibility: :exclamation:
OpenApiBuilderCustomizer - `customise()` -> `customize()` (here we cannot use the same trick as in `OpenApiCustomiser`, because it's the same class)
- Please confirm that you are fine about it, if not, this controversal change may be reverted.


springdoc-openapi-common/src/main/java/org/springdoc/core/GroupedOpenApi.java
```diff
@@ -197,13 +200,23 @@ public String getGroup() {
 		return pathsToExclude;
 	}
 
+	/**
+	 * Will be deleted in 2.0.
+	 */
+	@Deprecated
+	public List<OpenApiCustomiser> getOpenApiCustomisers() {
```
seregamorph, line 207: This code smells, but it gives BC.

springdoc-openapi-common/src/main/java/org/springdoc/core/SpringDocConfiguration.java
```diff
@@ -387,13 +387,13 @@ OperationCustomizer actuatorCustomizer() {
...
-		 * Actuator customizer OpenAPI customiser.
+		 * Actuator customizer OpenAPI customizer.
 		 *
-		 * @return the OpenAPI customiser
+		 * @return the OpenAPI customizer
 		 */
 		@Bean
 		@Lazy(false)
-		OpenApiCustomiser actuatorOpenApiCustomiser(WebEndpointProperties webEndpointProperties) {
+		OpenApiCustomizer actuatorOpenApiCustomizer(WebEndpointProperties webEndpointProperties) {
```
seregamorph, line 396: ❗may cause compatibility issues because of method renaming, e.g. if class is inherited (will raise compilation issue if `@Override` annotation is used)

springdoc-openapi-common/src/main/java/org/springdoc/core/customizers/ActuatorOpenApiCustomizer.java
```diff
@@ -17,29 +17,29 @@
 import static org.springframework.util.AntPathMatcher.DEFAULT_PATH_SEPARATOR;
 
 /**
- * The type Actuator open api customiser.
+ * The type Actuator open api customizer.
  * @author bnasslahsen
  */
-public class ActuatorOpenApiCustomizer implements OpenApiCustomiser {
+public class ActuatorOpenApiCustomizer implements OpenApiCustomizer {
```
seregamorph, line 23: hope this is fine (another interface)

springdoc-openapi-common/src/main/java/org/springdoc/core/customizers/OpenApiBuilderCustomizer.java
```diff
@@ -23,17 +23,17 @@
...
 public interface OpenApiBuilderCustomizer {
 
 	/**
-	 * Customise.
+	 * Customize.
 	 *
 	 * @param openApiService the open api builder
 	 */
-	void customise(OpenAPIService openApiService);
+	void customize(OpenAPIService openApiService);
```
seregamorph, line 37: ❗ pay attention: broken compatibility here

springdoc-openapi-common/src/main/java/org/springdoc/core/customizers/OpenApiCustomiser.java
```diff
@@ -36,4 +41,9 @@
 	 */
 	void customise(OpenAPI openApi);
 
+	@Override
+	default void customize(OpenAPI openApi) {
+		customise(openApi);
```
seregamorph, line 46: tricky, but fair enough

springdoc-openapi-hateoas/src/main/java/org/springdoc/hateoas/SpringDocHateoasConfiguration.java
```diff
@@ -82,24 +82,24 @@ CollectionModelContentConverter collectionModelContentConverter(HateoasHalProvid
...
 	 * @param springDocConfigProperties the spring doc config properties
-	 * @return the open api customiser
+	 * @return the open api customizer
 	 * @see org.springframework.hateoas.mediatype.hal.Jackson2HalModule.HalLinkListSerializer#serialize(Links, JsonGenerator, SerializerProvider) org.springframework.hateoas.mediatype.hal.Jackson2HalModule.HalLinkListSerializer#serialize(Links, JsonGenerator, SerializerProvider)org.springframework.hateoas.mediatype.hal.Jackson2HalModule.HalLinkListSerializer#serialize(Links, JsonGenerator, SerializerProvider)org.springframework.hateoas.mediatype.hal.Jackson2HalModule.HalLinkListSerializer#serialize(Links, JsonGenerator, SerializerProvider)
 	 */
 	@Bean
 	@ConditionalOnMissingBean
 	@Lazy(false)
-	OpenApiCustomiser linksSchemaCustomiser(HateoasHalProvider halProvider, SpringDocConfigProperties springDocConfigProperties) {
+	OpenApiCustomizer linksSchemaCustomizer(HateoasHalProvider halProvider, SpringDocConfigProperties springDocConfigProperties) {
```
seregamorph, line 96: method renamed (may cause compatibility issues)

springdoc-openapi-hateoas/src/main/java/org/springdoc/hateoas/converters/OpenApiHateoasLinksCustomiser.java
```diff
@@ -1,49 +1,24 @@
...
  * @author bnasslahsen
  */
-public class OpenApiHateoasLinksCustomiser extends SpecFilter implements OpenApiCustomiser {
-
-	/**
-	 * The Spring doc config properties.
-	 */
-	private final SpringDocConfigProperties springDocConfigProperties;
+@Deprecated
+public class OpenApiHateoasLinksCustomiser extends OpenApiHateoasLinksCustomizer implements OpenApiCustomiser {
```
seregamorph, line 14: extracted to super-class with correct name

springdoc-openapi-hateoas/src/main/java/org/springdoc/hateoas/converters/OpenApiHateoasLinksCustomizer.java
```diff
@@ -0,0 +1,48 @@
...
+	 *
+	 * @param springDocConfigProperties the spring doc config properties
+	 */
+	public OpenApiHateoasLinksCustomizer(SpringDocConfigProperties springDocConfigProperties) {
+		this.springDocConfigProperties = springDocConfigProperties;
+	}
+
+	@Override
+	public void customize(OpenAPI openApi) {
+		ResolvedSchema resolvedLinkSchema = ModelConverters.getInstance()
```
seregamorph, line 38: code moved exactly as-is from `OpenApiHateoasLinksCustomiser`

seregamorph:
BTW, [CONTRIBUTING.adoc](CONTRIBUTING.adoc) does not declare anything about back compatibility requirements (I believe, they exist).

bnasslahsen:
@seregamorph, 

 Thank you for your proposal. 
 In fact, this is not a typo from an English perspective as both are correct! 
 But it's due, to the fact of different contributions and we haven't paid this difference in the naming naming during the first merge.

 I agree it's better to stick to the same naming rules. Especially that other customizers naming is with "z" letter.
 After being found, it doesn't seem to be achievable to keep the backward compatibility with this change.

 We will keep your PR Open, in order to schedule it for next major release.

bnasslahsen:
The impact of this PR, on existing applications will be huge.
As many applications are relying on `OpenApiCustomiser` naming. 
We will keep it as is.

PR #1002, opened 2020-12-29
Fix NPE when using management port and spring.application.name is set

### LINKED ISSUE #996: Minimal project with `use-management-port` set causes NPE on api-docs endpoint
**Describe the bug**
In a minimal example, use-management-port is generating NPEs when accessing v3/api-docs.
```
java.lang.NullPointerException: null
	at org.springdoc.core.ActuatorProvider.getApplicationPort(ActuatorProvider.java:185) ~[springdoc-openapi-common-1.5.2.jar:1.5.2]
	at org.springdoc.api.AbstractOpenApiResource.getActuatorURI(AbstractOpenApiResource.java:1061) ~[springdoc-openapi-common-1.5.2.jar:1.5.2]
	at org.springdoc.webmvc.api.OpenApiActuatorResource.getServerUrl(OpenApiActuatorResource.java:152) ~[springdoc-openapi-webmvc-core-1.5.2.jar:1.5.2]
	at org.springdoc.webmvc.api.OpenApiResource.calculateServerUrl(OpenApiResource.java:295) ~[springdoc-openapi-webmvc-core-1.5.2.jar:1.5.2]
	at org.springdoc.webmvc.api.OpenApiResource.openapiJson(OpenApiResource.java:171) ~[springdoc-openapi-webmvc-core-1.5.2.jar:1.5.2]
	at org.springdoc.webmvc.api.OpenApiActuatorResource.openapiJson(OpenApiActuatorResource.java:131) ~[springdoc-openapi-webmvc-core-1.5.2.jar:1.5.2]
```

This appears to be because the ActuatorProvider is trying to access a port from a null `applicationWebServer`. However the previous call's if statement in AbstractOpenApiResource might also be taking the wrong path. It could also be that my example is too minimal and I am missing some configuration.

**To Reproduce**
* Steps to reproduce the behavior:
* Created a simple project with one api endpoint, no groupings.
* Add Springdoc, v3/api-docs returns valid apis.
* Enable use-management-port, and set a management port.
* actuator/openapi produces NPE

What version of spring-boot you are using?
- 2.4.1
What modules and versions of springdoc-openapi are you using?
- 1.5.2

**Expected behavior**
- The same response I get when accessing without using the management port.

spartusch:
The previous code attempts to differentiate between the application and the management servers by comparing event.getApplicationContext().getId(). This ID does contain the value of the spring.application.name property though. This property defaults to "application" and thus the IDs default to "application" and "application:management". The previous code does look for these default values. But this means that ActuatorProvider isn't initialized properly when spring.application.name is not set to the default value but to the application's name.

This fix checks the server namespace instead.

Fixes #996

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/ActuatorProvider.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/ActuatorProvider.java
@@ -30,6 +30,7 @@ import org.springdoc.api.AbstractOpenApiResource;
 import org.springframework.boot.actuate.autoconfigure.endpoint.web.WebEndpointProperties;
 import org.springframework.boot.actuate.autoconfigure.web.server.ManagementServerProperties;
 import org.springframework.boot.autoconfigure.web.ServerProperties;
+import org.springframework.boot.web.context.WebServerApplicationContext;
 import org.springframework.boot.web.context.WebServerInitializedEvent;
 import org.springframework.boot.web.server.WebServer;
 import org.springframework.context.ApplicationContext;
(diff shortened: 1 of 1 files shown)


bnasslahsen:
Thank you @spartusch for your contribution.

PR #1012, opened 2021-01-13
Feature/forward query params when redirecting to swagger UI

kaibra:
This would allow to link to one of multiple api-groups in the swagger-ui using the redirect in SwaggerWelcome:

```
https://localhost:8443/swagger-ui.html?urls.primaryName=my-group-2
```

which would now redirect to 

```
https://localhost:8443/swagger-ui/index.html?configUrl=/v3/api-docs/swagger-config&urls.primaryName=my-group-2
```

fixes https://github.com/springdoc/springdoc-openapi/issues/1013


bnasslahsen:
@kaibra, 

PR #1027, opened 2021-01-20
Rework operation customizing to consider returned value

### LINKED ISSUE #1025: Custom operation filters
We have little library for api versioning that generates custom RequestCondition depending on special annotation. So we can write controllers like this:
```java
@RestController
public class Controller {
    @Versioner(till = 0)
    @GetMapping("/test")
    public String old() {
        return "old";
    }

    @Versioner(from = 1)
    @GetMapping("/test")
    public Integer other() {
        return 123;
    }
}
```
Then `GET /v0/test` returns `"old"`  and `GET /v1/test` returns `123`

Now we need openapi to produce 2 groups: group `v0` with `GET /test` endpoint that produces `String` and group `v1` with `GET /test` endpoint that produces `Integer`

Something like `addCustomFilter` in `GroupedOpenApi.Builder` where I can use `handlerMethod.getMethodAnnotation(Versioner.class)` to determine, if this method should be displayed in this group

- I have tried to filter paths in `OpenApiCustomiser`, but when it is called, there is already only one endpoint because the others have been overwritten [here](https://github.com/springdoc/springdoc-openapi/blob/master/springdoc-openapi-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java#L402) due to identical `operationPath` (`/test`)

romash1408:
Fixes #1025

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java
@@ -400,7 +400,9 @@ public abstract class AbstractOpenApiResource extends SpecFilter {
 			buildCallbacks(openAPI, methodAttributes, operation, apiCallbacks);
 
 			// allow for customisation
-			customiseOperation(operation, handlerMethod);
+			operation = customiseOperation(operation, handlerMethod);
+			if (operation == null)
+				continue;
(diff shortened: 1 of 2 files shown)


bnasslahsen:
@romash1408, 

This PR doesn't seem relevant.
Note that each PR, should contains at least related tests ...

PR #1079, opened 2021-02-26
enable OpenApiCustomizer when config group using application.properties

lmtoo:
When config OpenApiCustomizer bean, and config springdoc.group-configs, OpenApiCustomizer and OperationCustomizer don't call when a Grouped OpenAPI generate docs


bnasslahsen:
Hi @lmtoo, 

 For this case, instead of using `springdoc.group-configs`, you should use `GroupedOpenApi` bean where you can declare the `operationCustomizers` and `openApiCustomisers`, which should only be **in the group level**.

PR #1085, opened 2021-03-04
Respect the springdoc.cache.disabled setting for recalculating oauth2 redirect url

floroks:
This pull request causes the oauth2 redirect url to be recalculated when the springdoc.cache.disabled setting is true, and the url was relative to begin with.

When springdoc is reachable under multiple host-names (e.g. through proxies & direct), the authentication-redirect-url always contains the host that was generated the first time for that specific instance. This PR will recalculate the url every time, when it was originally relative, and the configuration setting to disable caching is set (same behaviour as the generated server url in the api-docs). A unit-test was also added, to ensure this behaviour.

Legal notices below, as per company policy (sorry):
The program was tested solely for our own use cases, which might differ from yours.

Florian Roks \<florian.roks@daimler.com\>, Daimler TSS GmbH
[Imprint](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)

Licensed under APL-2


bnasslahsen:
@floroks, 

 Thank you for your PR.
 I have done the merge and added the support of webflux as wel.

PR #1096, opened 2021-03-08
Fix certain `WebConversionServiceProvider#convert` calls by supplying contextual information

nathankooij:
This provides contextual information (such as `@DateTimeFormat`) to the `WebConversionService`, such that conversions can be performed correctly. Without passing the `TypeDescriptor` these default values cannot be converted otherwise, and OpenAPI generation will fail altogether.


bnasslahsen:
Thank you @nathankooij for your contribution.

PR #1104, opened 2021-03-13
switch from 'redirect:' prefix to ResponseEntity for redirecting UI index

### LINKED ISSUE #1100: Redirect from /swagger-ui.html to /swagger-ui/index.html?configUrl=/v3/api-docs/swagger-config can lead to HTTPS downgrade
**Describe the bug**
We are running a spring-boot application behind a load balancer that terminates HTTPS. When accessing https://my-application.url/swagger-ui.html it redirects to http://my-application.url/swagger-ui/index.html?configUrl=/v3/api-docs/swagger-config. This is caused by the fact that SwaggerWelcomeWebMvc#redirectToUi returns a String with `redirect:`
prefix. This causes the tomcat to return an absolute URL, using HTTP as protocol instead of HTTPS, in the HTTP location header.

**To Reproduce**
- What version of spring-boot you are using: 2.4.1
- What modules and versions of springdoc-openapi are you using: `springdoc-openapi-ui`, `springdoc-openapi-data-rest`

**Expected behavior**
The redirect respects the used protocol, in this case HTTPS. I guess a relative redirect could solve this problem.

**Additional context**
As a workaround you can either access https://my-application.url/swagger-ui/index.html?configUrl=/v3/api-docs/swagger-config directly or set `server.tomcat.use-relative-redirects=true` in your application.properties

d3ns0n:
fixes #1100

### DIFF
--- a/springdoc-openapi-ui/src/main/java/org/springdoc/webmvc/ui/SwaggerWelcomeActuator.java
+++ b/springdoc-openapi-ui/src/main/java/org/springdoc/webmvc/ui/SwaggerWelcomeActuator.java
@@ -12,6 +12,7 @@ import org.springdoc.core.SwaggerUiConfigProperties;
 import org.springframework.boot.actuate.autoconfigure.endpoint.web.WebEndpointProperties;
 import org.springframework.boot.actuate.endpoint.web.annotation.ControllerEndpoint;
 import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.ResponseBody;
 
@@ -54,7 +55,7 @@ public class SwaggerWelcomeActuator extends SwaggerWelcomeCommon {
 	@Operation(hidden = true)
 	@GetMapping(DEFAULT_PATH_SEPARATOR)
 	@Override
-	public String redirectToUi(HttpServletRequest request) {
+	public ResponseEntity<Void> redirectToUi(HttpServletRequest request) {
 		return super.redirectToUi(request);
 	}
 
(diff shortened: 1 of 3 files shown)


d3ns0n:
I would like to add some tests for this, but I didn't manage to succeed. The existing tests, that check the Location header e.g. `test.org.springdoc.ui.app1.SpringDocRedirectDefaultTest`, already check for a relative path but I don't understand why these tests worked before. I suspect that it has something to do with how `@SpringBootTest` is spinning up the application server during the tests. Feedback and Suggestions are welcome as I would really like to make sure that the Location header is set properly.

bnasslahsen:
@d3ns0n, 

 Thank for your contribution.
 I would prefer also the switch from `redirect:` prefix.

 The tricky issue you had reported may need better integration tests.
 
 As far as the existing tests are passing, i will accept PR.

d3ns0n:
I was wondering if HTTP 302 is the proper response code as springdoc-openapi-webflux-ui returns 307 for the same use case. I found this on MDN

> The only difference between 307 and 302 is that 307 guarantees that the method and the body will not be changed when the redirected request is made. With 302, some old clients were incorrectly changing the method to GET: the behavior with non-GET methods and 302 is then unpredictable on the Web, whereas the behavior with 307 is predictable. For GET requests, their behavior is identical. -- https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307

So it wouldn't really change anything in this case since we only allow GET requests, but maybe we should make the return code consistent. What do you think?

bnasslahsen:
@d3ns0n, 

 I had the same thoughts. I think http 302 is much better.
 About making the change consistent, i also identified this action, but if you can propose an additional PR for webflux as well, it will be great!

d3ns0n:
I will propose a PR later tonight

PR #1107, opened 2021-03-15
use HTTP 302 status for UI redirect

d3ns0n:
@bnasslahsen 

here is the PR for switching to 302 instead of 307 when using webflux-ui. This PR also contains some cleanups for affected test classes:
- remove never thrown Exceptions from throws list
- remove IntelliJ warning `Argument 'getResult.getResponseBody()' might be null` by asserting to not null
- switch from `org.junit.jupiter.api.Assertions.assertTrue` to `org.assertj.core.api.Assertions.assertThat` for better assertion error output


bnasslahsen:
@d3ns0n, 

 Great PR. This one had more impacts on tests !
 Thank for your help.
 We should release the new version v1.5.6, before the end of the week, which will ship your changes as well.

PR #1145, opened 2021-04-23
Readme update

debargharoy:
Update readme to better organize contents and add a table of contents.


bnasslahsen (approved):
Thank you @debargharoy for your contribution.

debargharoy:
I assume the PRs are now being merged with a merge commit.

I was thinking if we should only keep the `Squash and Merge` option for merging PRs. That way, we can keep the number of commits less on the master branch and not carry over multiple smaller commits from the feature branch as shown below 
image

PR #1153, opened 2021-04-28
Modify ResponseSupportConverter to resolve inner type

jagen31:
This makes it so, given a type `IgnoredWrapper<C>` that you're trying to resolve, the fields of `C` will be resolved with `AnnotatedType`s pointing to `C` as the parent, which wasn't happening before.

We are trying to apply properties to the parent schema of some fields, that's the use case for the change.  When the parent is wrapped in some ignored wrapper, this fails due to the current behavior.


bnasslahsen:
@jagen31, 

Thank for your contribution.
Could you please add a related test for your change ?

jagen31:
@bnasslahsen done

jagen31:
oh my. 1 moment.

jagen31:
(done now)

bnasslahsen (approved):
Thank you @jagen31 for your contribution.

PR #1156, opened 2021-05-06
Pick up ExceptionHandlers from any parent of a class marked as @ControllerAdvice

eshishkin:
Currently the library scans  for exception handlers only declared method of `@ControllerAdvice`s. So, if an advice has parent classes with exception handlers, handlers will not be picked up and models will not be displayed in OpenAPI spec.

Sometimes, it might be useful to have one class with common handlers (e.g. a jar library that should be imported in all services across a project/company) and extend with more specific handlers.

Something like that

```java
-----------------------------------------
public class Error {
  // ...
}
-----------------------------------------
public class CommonErrorHandler {

  @ResponseStatus(INTERNAL_SERVER_ERROR)
  @ExceptionHandler(Exception.class)
  public Error onException(Exception ex) {
    return new Error();
  }

  // other common handlers
}
-----------------------------------------
@ControllerAdvice
public class ErrorHandler extends CommonErrorHandler {
  
  //additional error handlers

  @ResponseStatus(BAD_REQUEST)
  @ExceptionHandler(IllegalArgumentException.class)
  public SpecificError onIllegalArgumentException(IllegalArgumentException ex) {
    return new SpecificError ();
  }

}
-----------------------------------------
```

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/GenericResponseService.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/GenericResponseService.java
@@ -52,6 +52,7 @@ import org.springframework.core.ResolvableType;
 import org.springframework.core.annotation.AnnotatedElementUtils;
 import org.springframework.http.HttpStatus;
 import org.springframework.util.CollectionUtils;
+import org.springframework.util.ReflectionUtils;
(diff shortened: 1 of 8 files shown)


bnasslahsen:
Thank you @eshishkin  for your contribution.

Issue #1183, opened 2021-06-08
springdoc-openapi-common: Parameter reference is resolved as query parameter in generated schema


sadam21:
**Describe the bug**
I'm unable to add reference to `components/parameters`. Fields `in` with `query` value and String `schema` is automatically added by `OpenApiResource` despite these fields are properly defined in references parameter.

**To Reproduce**
I wanted to create reference to parameter as described in [OpenAPI specification](https://swagger.io/docs/specification/describing-parameters/). We have our implementation of `RepositoryRestResourceProvider` which creates parameter (io.swagger.v3.oas.models.parameters.Parameter)
```java
new Parameter.$ref("someReferencedParameter")
```
(`$ref` is the only parameter).

We also use `OpenApiResource` and the problematic code seems to be in [AbstractOpenApiResource::calculatePath](https://github.com/springdoc/springdoc-openapi/blob/master/springdoc-openapi-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java#L502) which contains:
```java
if (!CollectionUtils.isEmpty(operation.getParameters()))
    operation.getParameters().forEach(parameter -> {
        if (parameter.getSchema() == null)
            parameter.setSchema(new StringSchema());
        if (parameter.getIn() == null)
            parameter.setIn(ParameterIn.QUERY.toString());
    }
);
```
`query` value and string schema is added despite it's just reference where are properly defined different values.

**Expected behavior**
Generated parameter in OpenAPI schema should be:
```json
{"$ref":"#/components/parameters/someReferencedParameter"}
```

but was:
```json
{"$ref":"#/components/parameters/someReferencedParameter","in":"query","schema":{"type":"string"}}
```

**Additional context**
Used Springdoc version is 1.5.0. Version 1.5.9 seems to have the same mechanism according to the source code. I tried to upgrade but there are some incompatible changes which will take some time.

I'm not sure if it's bug or feature request. Is this behaviour intended? We used workaround with subclass of `Parameter`:

```kotlin
    private class RefParameter(referenceName: String) : Parameter() {
        init {
            `$ref` = referenceName
        }
        override fun getIn(): String? {
            return null
        }

        override fun getSchema(): Schema<*>? {
            return null
        }
    }
```

Is there better solution?

Thank you!

bnasslahsen:
Hi @sadam21, 

It is an enhancement request, to get it out of the box.
Note that you can still achieve your goal using `OpenApiCustomiser`.

[...]

PR #1197, opened 2021-06-21
Do not add default values for parameter references

sadam21:
Do not set default schema and "in" value for parameters which have
reference "$ref" set to not empty value.

https://swagger.io/docs/specification/describing-parameters/


sadam21:
This PR implements https://github.com/springdoc/springdoc-openapi/issues/1183.

bnasslahsen:
Thank you @sadam21 for your PR.

sadam21:
Thank you too @bnasslahsen , you're welcome!

PR #1256, opened 2021-09-03
Parent pom.xml as BOM

immens:
Giving ability to import springdoc-openapi as BOM


bnasslahsen:
Thank you @immens for your proposal.

Issue #1258, opened 2021-09-04
Add support for HandlerTypePredicate in spring-web


amiraziz:
**Describe the feature**

in my project, i need to add prefix for all @RestController beans mapping. so i add this config class:

```
@Configuration
public class AppConfiguration implements WebMvcConfigurer {

    @Override
    public void configurePathMatch(PathMatchConfigurer configurer) {
        configurer.addPathPrefix("/rest",
                HandlerTypePredicate.forAnnotation(RestController.class));
    }
}
```
after that i see swagger not preview correctly with default address, so I add this config keys:

```
springdoc.swagger-ui.path= swagger
springdoc.swagger-ui.url= /rest/api-docs
```
now i see this page with correct controller mapping but when i try it, url add prefix again and address change to `/rest/rest/...`

![image](https://user-images.githubusercontent.com/5073250/132086007-202b5dc3-5d60-4087-a7ce-80b8e96761fb.png)

```
@RestController
public class WorkflowController {

    private final WorkflowService workflowService;
    private final Environment environment;


    public WorkflowController(WorkflowService workflowService, Environment environment) {
        this.workflowService = workflowService;
        this.environment = environment;
    }

    @PostMapping(value = "/phoneCharge")
    public GeneralResponse startChargeProcess(@Valid @RequestBody MerchantChargeRequestDto dto, HttpServletRequest request) {
        return workflowService.startChargeProcess(dto, request.getRemoteAddr(), "ChargeProcess");
    }
}

```

I use this versions:

```
spring-boot-starter-parent => 2.3.12.RELEASE
springdoc-openapi-data-rest & springdoc-openapi-ui => 1.5.9
```

bnasslahsen:
The support has been added.
I have added a fix that you can test using the [latest available snapshot v1.5.11-SNAPHOT](https://springdoc.org/#how-can-i-use-the-last-springdoc-openapi-snapshot).

PR #1282, opened 2021-10-04
Use `ObjectProvider` to inject `RequestMappingInfoHandlerMapping`

hisener:
We noticed failures in some Spring Boot tests in the Springdoc 1.5.11 upgrade PR.

In [v1.5.11](https://github.com/springdoc/springdoc-openapi/releases/tag/v1.5.11), `SwaggerWelcomeWebMvc` injects an optional `RequestMappingInfoHandlerMapping`, however it's possible to have multiple beans for that type. For instance, it's not uncommon to have [`ControllerEndpointHandlerMapping`](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/web/reactive/ControllerEndpointHandlerMapping.java) in addition to `RequestMappingHandlerMapping`.

This change uses `ObjectProvider` instead so that we can handle multiple beans.

The exception we get:

`Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'swaggerWelcome' defined in class path resource [org/springdoc/webmvc/ui/SwaggerConfig.class]: Unsatisfied dependency expressed through method 'swaggerWelcome' parameter 3; nested exception is org.springframework.beans.factory.NoUniqueBeanDefinitionException: No qualifying bean of type 'org.springframework.web.servlet.mvc.method.RequestMappingInfoHandlerMapping' available: expected single matching bean but found 3: requestMappingHandlerMapping,webEndpointServletHandlerMapping,controllerEndpointHandlerMapping`

Relates to https://github.com/springdoc/springdoc-openapi/issues/1258 and https://github.com/springdoc/springdoc-openapi/commit/6e709c9da34f52087c75b9f61b33ad51093030e3.

### DIFF
--- a/springdoc-openapi-ui/src/main/java/org/springdoc/webmvc/ui/SwaggerConfig.java
+++ b/springdoc-openapi-ui/src/main/java/org/springdoc/webmvc/ui/SwaggerConfig.java
@@ -31,6 +31,7 @@ import org.springdoc.core.SwaggerUiConfigParameters;
 import org.springdoc.core.SwaggerUiConfigProperties;
 import org.springdoc.core.SwaggerUiOAuthProperties;
 
+import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.actuate.autoconfigure.endpoint.web.WebEndpointProperties;
 import org.springframework.boot.actuate.autoconfigure.web.server.ConditionalOnManagementPort;
 import org.springframework.boot.actuate.autoconfigure.web.server.ManagementPortType;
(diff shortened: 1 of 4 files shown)


shami13:
Hi @hisener 
Do you know some workaround for this  issue, until it not merged and released new version?
Regards, Mykhailo

hisener:
I think you could provide a `SwaggerWelcomeWebMvc` bean. So that it won't create the one in `SwaggerConfig` since it has `@ConditionalOnMissingBean` annotation.

We decided to stick with `1.5.10` until a fix is merged and released.

PR #1289, opened 2021-10-08
Fix server address customizing in case of enabled pre-loading mode

sokomishalov:


sokomishalov:
@bnasslahsen added test to cover the fix. But actually fix was even easier - we just need to check mutated new servers to be present instead of older ones.

PR #1290, opened 2021-10-08
Use Publisher as response wrapper to ignore instead of Mono

sokomishalov:
More abstract way to cover wrappers since spring supports any Publisher implementation.
My main concern is that Flux is also Publisher implementation and if it could break something.


bnasslahsen:
@sokomishalov, 

Your change is creating many regressions.
Please make sure, you add a test related to your code changes as well.

PR #1356, opened 2021-12-01
Add support for caching based on Locale

### LINKED ISSUE #1354: Add support for caching based on Locale.
**Describe the feature**
Caching does not seem to take locale into account.

https://github.com/springdoc/springdoc-openapi/blob/f86c3256594dca2f6e11ee67c2c71a7da8be4134/springdoc-openapi-common/src/main/java/org/springdoc/core/OpenAPIService.java#L727-L729

This can causes the wrong language to be displayed.

**To Reproduce**
Turn on caching (`springdoc.cache.disabled=false`)
Generate first request to swagger page using English locale
Generate second request swagger page using French locale

**Expected behavior**
I would expect that the second request using the French locale would return the French content however it returns the cached data which is the English content.

Note: Doing the reverse produces the opposite results. i.e. French local first and then English locale second will end up producing French data for the English locale.

**Additional context**
Work around seems to be to set 
`springdoc.cache.disabled=true`

ianwallen:
Add support for caching based on Locale. - Fixed #1354


bnasslahsen:
@ianwallen, 

Thank you @ianwallen for your efforts.
Tests are still failing for your PR.

Once corrected, i will proceed with the merge.

ianwallen:
Finally got it working - sorry for all the commits - I was having problems running all the test on my local system.

ianwallen:
@bnasslahsen 
I should point out that the cache that I implemented is based on the locale language only.
Looking at the following, this morning it seems like the messages files can get more complicated that simply using "en" or "fr"

https://sites.google.com/site/openjdklocale/design-notes/resource-bundle-lookup-order

So if someone is using different message files for en_US and en_UK, they would both be using the same "en" language cache so it would just load the first occurrence into the cache map.

I was originally going to use the Locale in the map however I was not sure if the Locale would work correctly during the lookup and after finding the following it seems like it could get more complicated that I wanted to go.

http://www.java2s.com/Code/Java/Collections-Data-Structure/MapusingLocaleobjectsaskeys.htm

Not sure if you want to leave it as is or flag this as a future enhancement?

bnasslahsen:
@ianwallen, 

I would say that if you feel confident about this enhancement, you can submit another PR.
I guess, this new PR should be easier than the original one!

PR #1359, opened 2021-12-03
Modify locale cache to support  country variants.

ianwallen:
Add locale cache support for country variants by using the language tag so that it can support messages from differences countries.  i.e. en_GB vs en_US

This is a correction to the previous pr #1356 which fixed issue #1354


bnasslahsen:
Thank you @ianwallen for your enhancement !

PR #1409, opened 2021-12-27
cleanup: chmod text files from 755 to 644

KENNYSOFT:
.java and .json files should not be executable. I've run the following commands in Windows and found out files with wrong permission.

```shell
git rm --cached -r .
git add .
```


bnasslahsen:
Thank you @KENNYSOFT for your contribution!

PR #1414, opened 2021-12-29
Fix dep class

mosesonline:
use interface instead of deprecated class


bnasslahsen:
@mosesonline, 

The problem of this change is it breaks compatibility with some legacy applications based on spring 4 / or spring-boot 1

bnasslahsen:
@mosesonline, 

I have done the merge making sure the compatibility with the spring 4 applications is not broken.
I am not convinced it's cleaner now.  But at least there is not more usage of the deprecated class.

mosesonline:
> @mosesonline,
> 
> I have done the merge making sure the compatibility with the spring 4 applications is not broken. I am not convinced it's cleaner now. But at least there is not more usage of the deprecated class.

ok, I didn't thought that you have to maintain compatibility to spring 4 (sb 1).

PR #1429, opened 2022-01-07
Allow excluding individual methods from OpenApi output

mc1arke:
It is currently possible to prevent methods from being included in the
OpenAPI output by ensuring the parent class is not covered by the
`packagesToScan` or `pathsToMatch` configuration, but where a method
that a user doesn't want included in the output is present in a class
that is included in the configuration, and that method shares a path
with other endpoints that also require inclusion, there's no way for a
user to exclude the method from being exposed as an operation in the
OpenAPI output.

To overcome this limitation, a MethodFilter has been introduced which
allows for a user to selectively exclude individual methods from being
parsed for their definitions. Every method that's detected for
potential inclusion in the OpenApi output is passed to the filter, and
any method that the filter rejects is excluded from further processing.
As multiple filters can be applied in a single configuration, the
result of all the filters is combined in an 'and' result, so all
filters must accept a method the have it available in the output.

This functionality allows for the annotation filtering feature in
SpringFox to be re-implemented by users, by creating a method filter
similar to `method -> method.isAnnotationPresent(MyAnnotation.class)`,
although allows much finer control since any other reflective
attributes of the method can also be checked at the same time.

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java
@@ -650,9 +650,26 @@ public abstract class AbstractOpenApiResource extends SpecFilter {
 	 * @return the boolean
 	 */
 	protected boolean isFilterCondition(HandlerMethod handlerMethod, String operationPath, String[] produces, String[] consumes, String[] headers) {
-		return isPackageToScan(handlerMethod.getBeanType().getPackage())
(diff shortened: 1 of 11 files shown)


bnasslahsen:
Thank you @mc1arke for your contribution to the project.

I have made little adjustments so the feature can also be available without using groups.
I have also renamed the interface `MethodFilter` to `OpenApiMethodFilter`.

As this feature will be available only programmatically (without using properties), i had to move some code from `SpringDocConfigProperties`.

PR #1430, opened 2022-01-07
springdoc-openapi-security: support custom login processing endpoints

lipniak:
My application uses a reverse proxy with only `/api/**` routes being available to the Spring server. Unfortunately, the login processing endpoint path in springdoc-openapi-security currently is fixed to `/login` so I had to make some changes in the code.
The new code uses the `requiresAuthenticationRequestMatcher` field of `usernamePasswordAuthenticationFilter` (it's private on `AbstractAuthenticationProcessingFilter` with no getter so we need to use the Reflection API) to find the login processing endpoint path.
I supported only `AntPathRequestMatcher` matchers because it is used by all default Spring Security filters when configuring with `HttpSecurity`. This should cover most use cases. 
I also have added a test `app8` that verifies if a login processing URL set as `/api/login` is properly reflected in API schema.

### DIFF
--- a/springdoc-openapi-security/src/main/java/org/springdoc/security/SpringDocSecurityConfiguration.java
+++ b/springdoc-openapi-security/src/main/java/org/springdoc/security/SpringDocSecurityConfiguration.java
@@ -20,6 +20,7 @@
 
 package org.springdoc.security;
 
+import java.lang.reflect.Field;
 import java.util.Optional;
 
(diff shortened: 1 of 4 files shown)


bnasslahsen:
Thank you @lipniak for your contribution!

ABoat365:
@lipniak Can support webflux?

Issue #1469, opened 2022-01-26
CSRF header should not be sent to cross domain sites


sebastien-helbert:
**Describe the bug**

When CSRF is enabled a CSRF token is sent for each issued request including to cross domain sites.

This is a problem when the third party server does not allow this header to be sent (This use case is well described in https://github.com/springdoc/springdoc-openapi/issues/1036)
and this potentially may be exploited by an attacker as it can allow a third party site to obtain a token from the main site and therefore forge a CSRF request.

**To Reproduce**

This can be reproduced easily with a password OAuth flow security scheme and a keycloak server (which by default does not authorize the x-xsrf-token header to be sent because it is not expected). See https://github.com/springdoc/springdoc-openapi/issues/1036 

**Expected behavior**
The CSRF header should not be added for cross domain sites

bnasslahsen:
@sebastien-helbert ,

Do not create duplicates. https://github.com/springdoc/springdoc-openapi/issues/1228

sebastien-helbert:
this is not a duplicate

sebastien-helbert:
> @sebastien-helbert ,
> 
> Do not create duplicates. #1228

this is not a duplicate

bnasslahsen:
> > @sebastien-helbert ,
> > Do not create duplicates. #1228
> 
> this is not a duplicate

Ok 👍

Issue #1228, opened 2021-07-28
Add option to get the CSRF token from the Local Storage


esfomeado:
**Is your feature request related to a problem? Please describe.**
Allow to get the CSRF token from the Local Storage and not only from the cookies.

**Describe the solution you'd like**
Add a new configuration to load the CSRF token from the Local Storage.

**Describe alternatives you've considered**
The alternative solution it's to write a custom request interceptor.

bnasslahsen:
@esfomeado, 

Are you willing to propose a PR for this feature ?

esfomeado:
Yeah, I can do this when I have some time.

bnasslahsen:
ok. I am closing this ticket.
You can directly submit your PR and will be glad to review it.

Issue #1036, opened 2021-01-26
Swagger UI: X-XSRF-TOKEN is sent to OAuth2 token url


daniel-shuy:
**Describe the bug**
When configuring OpenAPI with OAuth2 security scheme and springdoc-openapi-ui with `springdoc.csrf.enabled=true`, Swagger UI will include `X-XSRF-TOKEN` in the request to the OAuth2 token url.

Certain OAuth2 providers (in my case, Keycloak) with CORS protection may cause the browser to block the request due to the presence of `X-XSRF-TOKEN`, since it is not typically in the `Access-Control-Allow-Headers` list.

**To Reproduce**
- What version of spring-boot you are using? 2.4.1
- What modules and versions of springdoc-openapi are you using? springdoc-openapi-ui:1.5.1
- Provide with a sample code (HelloController) or Test that reproduces the problem:
```java
new OpenAPI()
    .addSecurityItem(new SecurityRequirement()
        .addList("keycloak")
    )
    .components(new Components()
        .addSecuritySchemes("keycloak", new SecurityScheme()
            .type(SecurityScheme.Type.OAUTH2)
            .flows(new OAuthFlows()
                .authorizationUrl("http://<keycloakUrl>/auth/realms/<realm>/protocol/openid-connect/auth")
                .tokenUrl("http://<keycloakUrl>/auth/realms/<realm>/protocol/openid-connect/token")
            )
        )
    );
```
```yml
springdoc:
  swagger-ui:
    csrf:
      enabled: true
```

**Expected behavior**
Swagger UI should not include `X-XSRF-TOKEN` for OAuth2 token url requests.

**Screenshots**

**Additional context**

bnasslahsen:
@daniel-shuy, 
  
 `X-XSRF-TOKEN` is the default value for many providers.
  You can adapt the `cookie-name` and `header-name, as described in the [swagger-ui-properties section 
](https://springdoc.org/#swagger-ui-properties)

daniel-shuy:
@bnasslahsen what I meant is that the `X-XSRF-TOKEN` should still be sent in requests to the endpoint urls, but excluded from the requests to the OAuth2 token url.

The CSRF support is meant for the application, not for the OAuth2 provider. My endpoints (using Spring Security) still expects `X-XSRF-TOKEN`, but Keycloak does not expect any CSRF token (there are no CSRF headers in Keycloak's `Access-Control-Allow-Headers`, renaming the header name does not solve this issue).

[...]

PR #1470, opened 2022-01-26
CSRF header should not be sent to cross domain sites #1469

sebastien-helbert:
fix for #1469 and #1036
not a duplicate of #1228


bnasslahsen:
Thank you @sebastien-helbert for your contribution

GandalfTheBlack16:
There were some backslashes missing into addCSRFLocalStorage method that was producing a HTML syntax error at runtime. 
Solved #1501

Issue #1473, opened 2022-01-26
JacksonFeature NoClassDefFoundError


youming-lin:
**Describe the bug**
I'm getting error running unit tests:
```
java.lang.NoClassDefFoundError: com.fasterxml.jackson.core.util.JacksonFeature
	at com.fasterxml.jackson.databind.ObjectMapper.<init>(ObjectMapper.java:673)
	at com.fasterxml.jackson.databind.ObjectMapper.<init>(ObjectMapper.java:576)
	at io.swagger.v3.core.util.ObjectMapperFactory.create(ObjectMapperFactory.java:79)
	at io.swagger.v3.core.util.ObjectMapperFactory.createJson(ObjectMapperFactory.java:65)
	at io.swagger.v3.core.util.Json.mapper(Json.java:13)
	at org.springdoc.ui.AbstractSwaggerIndexTransformer.<init>(AbstractSwaggerIndexTransformer.java:86)
	at org.springdoc.webmvc.ui.SwaggerIndexPageTransformer.<init>(SwaggerIndexPageTransformer.java:56)
<rest omitted>
```

I believe the issue is due to the `jackson-core` in `webjars-locator-core` is overriding its version for everything else:
<img width="464" alt="Screen Shot 2022-01-26 at 11 17 32 AM" src="https://user-images.githubusercontent.com/19825876/151214360-09fe5756-00ce-4e31-b7c4-ea3e60a6d6f2.png">

Per https://fasterxml.github.io/jackson-core/javadoc/2.12/com/fasterxml/jackson/core/util/JacksonFeature.html, `JacksonFeature` is added in 2.12.0, which is newer than the version in `webjars-locator-core`.

Fix is to update `webjars-locator-core` to 0.48.0.

PR #1474, opened 2022-01-26
update webjars-locator-core version

youming-lin:
Fixes https://github.com/springdoc/springdoc-openapi/issues/1473


bnasslahsen:
Thank you @youming-lin for tour contribution !

Issue #1476, opened 2022-01-27
Whether the project structure can be adjusted


Hccake:
First of all, I'm really sorry that I opened another issue. The last issue I opened, issue #1450, was marked as a duplicate, but it wasn't actually a duplicate, and I can't reply right now.

I just want to discuss whether **springdoc-openapi-common** can be split into smaller and simpler modules, such as adding a **springdoc-openapi-core** module to migrate some necessary annotations and models from common?
Or move the `spring.factories` file to other modules, so that **springdoc-openapi-common** is no longer automatically configured in spring-boot？

For example, there are the following module divisions：
```
|-- common-model			
|-- server-a		
|-- server-b
```
Some query objects are defined in the common-model module, using the `@ParameterObject` annotation, so it introduces the springdoc-openapi-common dependency.

server-b is a spring-boot project that does not require external openapi documentation, but since it depends on common-model, we have to add springdoc shutdown configuration for it. Even when upgrading the springdoc version, there will be previous exceptions, like #1446.

Based on the above scenario, I hope **springdoc-openapi-common** can make some adjustments, 

thank you

bnasslahsen:
@Hccake, 

There is [ongoing enhancements](https://github.com/springdoc/springdoc-openapi/tree/2.x).
`springdoc-openapi-common` will be renamed to `springdoc-openapi-starter-common` to be more explicit.
But if you want disable it. Just exclude the related auto-configuration classes located in the `spring.factories`.

If you have better enhancement and feel free to propose a PR (Just make sure there is no duplication in your PR!).

Hccake:
@bnasslahsen 

Thanks for your reply, I tried making a little change on the master branch and opened a PR.

Just added a new module and moved some files to this module, without changing any package names for compatibility.

Please review when you have time

PR #1478, opened 2022-01-27
Split out a springdoc-openapi-core module from springdoc-openapi-common

Hccake:
close gh-1476


bnasslahsen:
@Hccake, 

This is going to be huge change, and adding more complexity to the project. 

We don't really need an additional module for this.
You can just rely on Exclude Auto-Configuration Classes `@EnableAutoConfiguration(exclude=YourClasses.class)` which is most relevant in this case.

Hccake:
I provide some basic dependencies to third parties, which contain some entity classes and use the annotations provided by OpenAPI.

In the current design, 3rd parties using the dependencies I provided must also actively disable this auto-configuration, even if they don't want to use OpenAPI at all，so, i think it is necessary to split the functional class and the configuration class

PR #1480, opened 2022-01-28
Fix the problem that the inconsistent newline characters of different platforms

Hccake:
The newline character of linux system is LF,
The default newline character in Windows system files is CRLF.

When git clones the code, it will automatically do a newline conversion for different platforms, which causes the newline of the tested file to be CRLF in the windows system, which does not match the expected return value.

In addition, because the CR character is used in three places in swagger-ui/index.html, and it is just before the carriage return character LF, which will cause the original CR character representation to be lost when git clone converts the newline character, so in order to cross The unification of the platform needs to ignore the comparison of CR characters when judging


bnasslahsen:
@Hccake, 

Thank you for you contribution to the project.

PR #1500, opened 2022-02-09
Pick up exception handler in case there is no controller advice at alll, fixes 1498

### LINKED ISSUE #1498: Support for @ExceptionHandler inside @RestController only works if a @RestControllerAdvice class is present
**Describe the bug**
The @ExceptionHandler methods inside a @RestController are picked up correctly if and only if there is at least one class present which is annotated with @RestControllerAdvice. This class might be empty, though.

**To Reproduce**
I created a fork of the master branch and added a new test app182 which is a copy of the test app124.
Except that both @ExceptionHandlers are contained in the @RestController and the @RestControllerAdvice class was removed.

https://github.com/SimonAlthoff/springdoc-openapi/commit/8db9112a0be42ad41f9f3c6a8844570b0fc43cdf

**Expected behavior**
@ExceptionHandler within a @RestController are picked up even when no @RestControllerAdvice class exists.

**Additional context**
The original feature was implemented in ticket #748

eshishkin:
This PR fixes #1498 issue. 
Exception handlers can be resolved even if no @ControllerAdvice is registered


bnasslahsen:
Thank you @eshishkin for your contribution !

PR #1514, opened 2022-02-19
Added support for kotlinx.serialization

werner77:
This PR adds support for kotlinx.serialization via an implementation of ModelConverter.

### DIFF
--- a/springdoc-openapi-kotlin/pom.xml
+++ b/springdoc-openapi-kotlin/pom.xml
@@ -31,6 +31,12 @@
 			<scope>provided</scope>
 		</dependency>
 		<dependency>
+			<groupId>org.jetbrains.kotlinx</groupId>
+			<artifactId>kotlinx-serialization-core-jvm</artifactId>
+			<version>1.3.2</version>
+			<scope>provided</scope>
+		</dependency>
+		<dependency>
 			<groupId>org.springdoc</groupId>
 			<artifactId>springdoc-openapi-webflux-core</artifactId>
 			<version>${project.version}</version>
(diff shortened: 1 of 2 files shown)


JesusMcCloud:
I might be doing things wrong, but when I added your code and explicitly calling `    ModelConverters.getInstance().addConverter(KotlinxSerializationTypeConverter(Json, false))` before `runApplication`, I still only get the default, messed-up Jackson serialization in Swagger-UI (and in the generated OpenAPI JSON), which obviously entails that custom serializers and`@SerialName`annotations are ignored as well…

How can I debug this? How can I help?

JesusMcCloud:
The [last commit](https://github.com/springdoc/springdoc-openapi/pull/1514/commits/e2ee75607753c7cb054efa3d3ceb88ba582587c1) helped, but it's quite not there yet:

For example: we're using custom serializers for `X509Certificate` and it does show up correctly in the schemas. However, the request and responses, which contain certificates still use the default (faulty) serializers. Maybe the cause of this behaviour is related to `X509Certificate` being a platform type

werner77:
Yes right now it only considers types which are annotated with '@Serializable'. I can work on a fix for that.

JesusMcCloud:
Put differently, What's missing is properties annotated with `@Serializable(with=…`. Another Thing: Does your code dealt with the (experimental) [`@Serializer(forClass = …)`](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/serializers.md#external-serialization-uses-properties)  Annotation?

bnasslahsen:
@werner77, 

Have you finished your implementation ?
If it's the case. Could you also consider adding corresponding unit tests to the project?

werner77:
When I have time (hopefully soon) I will fix the open issues and add some unit tests.

bnasslahsen:
closed as no activity

PR #1543, opened 2022-03-09
Add ServerBaseUrlCustomizer functionality to be able to customize the generated Server URL

skylar-stark:
ServerBaseUrlCustomiser to allow making changes to the generated server base URL. Also adds a getter method for retrieving the generated server base URL to use in an OpenApiBuilderCustomizer in order to make customizations based on the generated value

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/OpenAPIService.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/OpenAPIService.java
@@ -62,7 +62,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springdoc.core.customizers.OpenApiBuilderCustomizer;
-
+import org.springdoc.core.customizers.ServerBaseUrlCustomizer;
 import org.springframework.beans.BeansException;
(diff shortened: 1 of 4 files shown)


skylar-stark:
Use case: strip the "context-path" off the generated server URL and instead add it to the beginning of each path, so that Open API generation works behind a reverse proxy

bnasslahsen:
@skylar-stark, 

Could you add the corresponding unit tests to your PR?

skylar-stark:
@bnasslahsen Unit test added. Thanks!

bnasslahsen:
@skylar-stark, 

Thank you for your contribution !

springdoc-openapi-common/src/main/java/org/springdoc/core/OpenAPIService.java
```diff
@@ -120,6 +120,11 @@ public class OpenAPIService implements ApplicationContextAware {
 	 */
 	private final Optional<List<OpenApiBuilderCustomizer>> openApiBuilderCustomisers;
 
+	/**
+	 * The server base URL customisers.
+	 */
+	private final Optional<List<ServerBaseUrlCustomizer>> serverBaseUrlCustomisers;
```
JoshMcCullough, line 126: Weird pattern to use `Optional` for class vars when the code has to check if it's null anyway.

PR #1579, opened 2022-03-28
Updated class and method javadoc handling

gebezs:
- [X] dto's class javadoc:

  * added as description for the schema

- [X] controller's class javadoc:

  * added as description of the tag

- [X] controller's method javadoc:

  * first line as summary
  * everything as description

- [X] filters the generic responses by the declared exceptions:

  * only if `override-with-generic-response-if-declared` is `true`

- [X] `@throws` in method javadoc:

  * overrides the `@return` of the `ExceptionHandler`'s method
  * only if `override-with-generic-response-if-declared` is `true`

- [x] Formatting
- [x] Test update for existing tests
  *  ~150 tests
  * Run them one by one and compare the result to make sure nothing else changed


bnasslahsen:
@gebezs, 

Thank you for your contribution!

PR #1593, opened 2022-04-05
Add-Support-For-Servlet-Paths

lbenedetto:
In my project, the documentation exists inside of a servlet with its own separate context path.
This fix (and only this fix) is required for everything to work as expected in my project.

Would appreciate it if something similar to this could be added.


bnasslahsen:
@lbenedetto, 

Your PR, is breaking a lot of tests.
Not that servlet paths are supported by just adding: `spring.mvc.servlet.path=/your-servlet-path`

Issue #1609, opened 2022-04-12
Customizer for all OpenAPI descriptions ("default" and GroupedOpenApi)


christophejan:
**Is your feature request related to a problem? Please describe.**

Currently, OpenApiCustomiser & OperationCustomizer beans are take in account only for the "default" OpenAPI description. When GroupedOpenApi are declared to just obtain a subset of the described operations, customizer beans are ignored and so the result is inconsistent.

When GroupedOpenApi are declared programmatically, it's possible to manually retrieve the customizer beans and apply them, but I think this Springdoc could make it easier.

When GroupedOpenApi are declared through spring properties, this is especially uncomfortable as I see no way to apply customizer on them.

I think the main usage for GroupedOpenApi is to build an OpenAPI description on a subset of available operations (i.e. a filtered view), and the current behavior on customizer don't seems to ease such use cases.

**Describe the solution you'd like**

I would like a way to specify customizers applied to all OpenAPI descriptions ("default" and GroupedOpenApi) (i.e. global).

This can be achieved by many ways :
1. Either by applying OpenApiCustomiser & OperationCustomizer beans on GroupedOpenApi ; but this could have quite strong impacts on existing applications
1. Or by introducing GlobalOpenApiCustomiser & GlobalOperationCustomizer interfaces
1. Or by introducing and using a GroupedOpenApiBuilder bean that can be either overridden by applications (@ConditionalOnMissingBean) or customized by providing bean(s) implementing some interface

Etc...

**Additional context**

Springdoc 1.6.7 - Kotlin webflux

bnasslahsen:
@christophejan; 

Groups and Global OpenAPI definitions are different.
You can define a single `OpenApiCustomiser` Bean which will be used by the Gloabl OpenAPI Bean. 
And for the `GroupedOpenApi` reference this `OpenApiCustomiser` using `addOpenApiCustomizer`.

christophejan:
@bnasslahsen, you have just described the behavior I've already given.

This was a feature request to improve current behavior ; I was expecting a path to improve springdoc handling of use case as simple as : some operations need customization whatever the requested OpenAPI description (default one or group).

I will try to rephrase the points to improve :
1. By default, /v3/api-doc/mygroup is inconsistent with /v3/api-docs for operations customized through OperationCustomizer
1. GroupedOpenApi declared through spring properties (yml) should be customizable
1. Make easier to apply all customizer beans on GroupedOpenApi declared programmatically

Regards

[...]

PR #1616, opened 2022-04-18
Add global customizer and filters

christophejan:
* #1609

Add  GlobalOpenApiCustomiser GlobalOperationCustomizer and GlobalOpenApiMethodFilter that will be applied to :
1. _Default_ OpenAPI description
2. GroupedOpenApi declared through spring properties (yml)
3. GroupedOpenApi declared programmatically like :
```
@Bean
public GroupedOpenApi mygroup(GroupedOpenApi.Builder builder) {
    return builder
        .group("mygroup")
        [...]
        .build();
}
```

The PR is composed by 4 commits that are more easy to review one by one. Only the last one add global customizer/filter. 

Here some details on the commits :
1. _Extra assertions on default group (on current behavior)_ :
I would **warn about a preexisting suspicious behavior** when ```springdoc.use-management-port=true```, querying nonexistent groups return a 404 with Webflux but a 500 with MVC.
1. _Declare the most specific return type possible on some bean declaration_ :
This is safer for components that implement several interfaces or for components potentially referred to by their implementation type (see [Declaring a bean](https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#beans-java-declaring-a-bean) tip on spring core documentation)
1. _Interact only with bean definition during bean registration phase_ :
This will prevent any premature bean instantiation, violating the container and causing unintended side-effects (see [BeanFactoryPostProcessor](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/config/BeanFactoryPostProcessor.html) documentation). In concrete terms :
    * register bean definition rather than bean instance
    * do not require GroupedOpenApi beans instances anymore for SpringdocActuatorBeanFactoryConfigurer bean registrations
1. _Add global customizers and filters_

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/SpringDocConfiguration.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/SpringDocConfiguration.java
@@ -51,7 +51,6 @@ import org.springdoc.core.customizers.DataRestDelegatingMethodParameterCustomize
 import org.springdoc.core.customizers.DelegatingMethodParameterCustomizer;
 import org.springdoc.core.customizers.OpenApiBuilderCustomizer;
 import org.springdoc.core.customizers.OpenApiCustomiser;
-import org.springdoc.core.customizers.OperationCustomizer;
 import org.springdoc.core.customizers.PropertyCustomizer;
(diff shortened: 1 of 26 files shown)


christophejan:
@bnasslahsen, thank you for all your hard work on Springdoc

PR #1630, opened 2022-04-26
remove repeated HttpSession

ghostg00:


bnasslahsen:
@ghostg00, 

Good catch !

PR #1687, opened 2022-06-01
fix(kotlin): properly generate ByteArray property

### LINKED ISSUE #1688: kotlin ByteArray property incorrect
**Describe the bug**
- If you are reporting a bug, please help to speed up problem diagnosis by providing as much information as possible:
- A clear and concise description of what the bug is: the title of an issue is not enough

When a model contains a ByteArray field, springdoc does not correctly generate a property definition.

currently this is generated

```json
"properties": {
          "data": {
            "type": "array",
            "items": {
              "type": "string",
              "format": "byte"
            }
          }
        }
```

When is should be 

```json
        "properties": {
          "data": {
            "type": "string",
            "format": "byte"
          }
        }
```

as per https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#data-types

**To Reproduce**

I've fixed it here already: https://github.com/springdoc/springdoc-openapi/pull/1687

I'm honestly happy to fix it myself, I just want to document the bug now so no one else gets stuck on it.

**Screenshots**

**Additional context**

codymikol:
this removes an incorrect wrapper around bytearray
properties.

Fixes #1688

### DIFF
--- a/springdoc-openapi-kotlin/src/main/java/org/springdoc/kotlin/SpringDocKotlinConfiguration.java
+++ b/springdoc-openapi-kotlin/src/main/java/org/springdoc/kotlin/SpringDocKotlinConfiguration.java
@@ -23,8 +23,10 @@
 package org.springdoc.kotlin;
 
 import com.fasterxml.jackson.module.kotlin.KotlinModule;
+import io.swagger.v3.oas.models.media.ByteArraySchema;
 import kotlin.Deprecated;
 import kotlin.coroutines.Continuation;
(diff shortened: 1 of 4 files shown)


bnasslahsen:
Thank you @codymikol for your contribution to the project.

codymikol:
No problem :) any idea when the next release is scheduled? Thanks for this project!

bnasslahsen:
@codymikol,

By the end of July!

codymikol:
Awesome! Thanks for the quick response!

PR #1691, opened 2022-06-01
Update ObjectMapperProvider.java

fabienfleureau:
fix: fix ObjectMapper configuration when writer-with-order-by-keys is set to true


bnasslahsen:
Finally fixed by  #1713!

PR #1700, opened 2022-06-08
fix: Support PageableDefault#value()

daniel-shuy:
[PageableDefault#value()](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/web/PageableDefault.html#value--) is an alias for [PageableDefault#size()](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/web/PageableDefault.html#sort--), but currently, `springdoc-openapi` is only reading the default size from `size`, not `value`.

This change attempts to get the default size from `size` first. If it is the default value, then it will get the default size from `value`. I'm not sure if this is the correct precedence, anyone knows if Java annotation parsers normally give higher precedence to `value` or the aliased parameter?


daniel-shuy:
Thanks @bnasslahsen!

I tested configuring both `size` and `value` for `PageableDefault` on the same parameter, and it looks like Spring gives `size` the higher priority, as expected.

PR #1702, opened 2022-06-08
Support Spring Data Sort and @SortDefault

### LINKED ISSUE #1668: Support Spring Data Sort
`springdoc-openapi` supports [Pageable](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/web/PageableDefault.html), so it make sense to support [Sort](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/domain/Sort.html) as well.

Seeing #1186 and #1618, and now needing it myself, there is quite a bit of demand for it.

@bnasslahsen I would like to contribute a PR for it, but would first like to consult you on what is the best approach.

From reading the existing implementation, since `Sort` is represented as a single `String[]` parameter, the easiest way would be to implement a `ModelConverter` that returns a `new ArraySchema().items(new StringSchema())`, e.g.
```java
	@Override
	public Schema resolve(AnnotatedType type, ModelConverterContext context, Iterator<ModelConverter> chain) {
		JavaType javaType = springDocObjectMapper.jsonMapper().constructType(type.getType());
		if (javaType != null) {
			Class<?> cls = javaType.getRawClass();
			if (SORT_TO_REPLACE.equals(cls.getCanonicalName())) {
				if (!type.isSchemaProperty())
					return new ArraySchema().items(new StringSchema());
				else
                                         // also, what's the purpose of this?
					type.name(cls.getSimpleName() + StringUtils.capitalize(type.getParent().getType()));
			}
		}
		return (chain.hasNext()) ? chain.next().resolve(type, context, chain) : null;
	}
```

However, I think it would be nice if the resolved parameter would have the same `description` as `Pageable#sort`.

Would it make sense to create a model class for only 1 parameter? e.g.
```java
public class Sort {
  @Parameter(description = "Sorting criteria in the format: property,(asc|desc). "
      + "Default sort order is ascending. " + "Multiple sort criteria are supported."
      , name = "sort"
      , array = @ArraySchema(schema = @Schema(type = "string")))
  private List<String> sort;

  public List<String> getSort() {
    return sort;
  }

  public void setSort(List<String> sort) {
    if (sort == null) {
      this.sort.clear();
    }
    else {
      this.sort = sort;
    }
  }
}
```

Or would it be better to create an anonymous class of `ArraySchema` and override `ArraySchema#getDescription()`? e.g.
```java
return new ArraySchema() {
  @Override
  public String getDescription() {
    return "Sorting criteria in the format: property,(asc|desc). Default sort order is ascending. Multiple sort criteria are supported.";
  }

  @Override
  public Schema<?> getItems() {
    return new StringSchema();
  }
};
```
or create a subclass of `ArraySchema`?
```java
public class SortSchema extends ArraySchema {
  private Schema<?> items = new StringSchema();
  private String description = "Sorting criteria in the format: property,(asc|desc). Default sort order is ascending. Multiple sort criteria are supported.";

  @Override
  public Schema<?> getItems() {
    return items;
  }

  @Override
  public void setItems(Schema<?> items) {
    this.items = items;
  }

  @Override
  public String getDescription() {
    return description;
  }

  @Override
  public void setDescription(String description) {
    this.description = description;
  }
}
```
but I don't see any such existing patterns, so I'm sure if its the right way to do it.

daniel-shuy:
Closes #1668

Support Spring Data [Sort](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/domain/Sort.html) and [@SortDefault](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/web/SortDefault.html)

image

I took the opportunity to refactor out the swagger annotation delegations from `DataRestDelegatingMethodParameterCustom` into separate classes, so that its clearer what annotation parameters are overridden.

Note that `@SortDefault` can be used on `Pageable` (not just `Sort`). If both `@PageableDefault` and `@SortDefault` are present on the same parameter, `@SortDefault` takes priority (I've implemented it to have the same behavior as Spring).

[@SortDefault#value](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/web/SortDefault.html#value--) is an alias for [@SortDefault#size](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/web/SortDefault.html#sort--). If both are configured, `sort` takes priority (I've implemented it to have the same behavior as Spring).

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/Constants.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/Constants.java
@@ -98,6 +98,11 @@ public final class Constants {
 	public static final String SPRINGDOC_POLYMORPHIC_CONVERTER_ENABLED = "springdoc.model-converters.polymorphic-converter.enabled";
 
 	/**
+	 * The constant SPRINGDOC_SORT_CONVERTER_ENABLED.
+	 */
+	public static final String SPRINGDOC_SORT_CONVERTER_ENABLED = "springdoc.model-converters.sort-converter.enabled";
+
+	/**
 	 * The constant SPRINGDOC_SCHEMA_RESOLVE_PROPERTIES.
 	 */
 	public static final String SPRINGDOC_SCHEMA_RESOLVE_PROPERTIES = "springdoc.api-docs.resolve-schema-properties";
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/SpringDocConfigProperties.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/SpringDocConfigProperties.java
@@ -717,6 +717,10 @@ public class SpringDocConfigProperties {
 		 */
 		private PolymorphicConverter polymorphicConverter = new PolymorphicConverter();
 
+		/**
+		 * The Sort converter.
+		 */
+		private SortConverter sortConverter = new SortConverter();
 
 		/**
 		 * Gets deprecating converter.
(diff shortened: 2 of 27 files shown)


bnasslahsen:
Thank you @daniel-shuy for your contribution.
I had to do the merge manually to resolve conflicts. 
Don't hesitate to let me know if there are any other enhancements.

shrralis:
@daniel-shuy @bnasslahsen it doesn't cover the `@SortDefault.SortDefaults`, does it?

PR #1718, opened 2022-06-22
fixes #1717 javadoc of JsonUnwrapped fields not set

### LINKED ISSUE #1717: javadoc of @JsonUnwrapped fields not set
- using springdoc-openapi-javadoc 1.6.9
- the controller contains a parameter or return type of type Foo
- Foo contains a field of type Bar which is annotated with @JsonUnwrapped
- the javadoc of the fields of Bar is not set to the output description

**To Reproduce**
download and unzip the sample project
[demo-springdoc-javadoc.zip](https://github.com/springdoc/springdoc-openapi/files/8956014/demo-springdoc-javadoc.zip)
run with 
mvn spring-boot:run
open
http://localhost:8080/swagger-ui/index.html
select Schema

**Expected behavior**
the javadoc of the Bar field should be set in the api-docs and hence will be displayed by swagger-ui

**Screenshots**
This is how it looks like
grafik
This is how it should be (generated with the proposed solution)
grafik

**Solution proposal**
in org.springdoc.openapi.javadoc.JavadocPropertyCustomizer.setJavadocDescription(Class<?>, List<Field>, Schema)
the method should be called recursively for all types of fields annotated with @JsonUnwrapped

`fields.stream().filter(f -> f.isAnnotationPresent(JsonUnwrapped.class))
                    .forEach(f -> setJavadocDescription(f.getType(), FieldUtils.getAllFieldsList(f.getType()), existingSchema));`

shardt68:


meiweimuli:
When will this modification be added to [springdoc-openapi-starter-common/src/main/java/org/springdoc/core/customizers/JavadocPropertyCustomizer.java]? Also, if the property decorated by the JsonUnwrapped annotation is generic, the type obtained by f.getType() will be Object, and the real fields cannot be obtained. Thank you. @bnasslahsen

PR #1742, opened 2022-07-15
Optimize the group order problem

zhangtongjob:
for example，The display on springdoc is unordered
```
 @Bean
  public GroupedOpenApi userApi(){
      return GroupedOpenApi.builder()
          .group("user")
          .pathsToMatch(/user/**")
          .build();
  }

  @Bean
  public GroupedOpenApi orderApi(){
      return GroupedOpenApi.builder()
          .group("order")
          .pathsToMatch("/order/**")
          .build();
  }
```


bnasslahsen:
Thank you @zhangtongjob for your contribution to the project!

Issue #1743, opened 2022-07-18
change code access level


NaccOll:
**Is your feature request related to a problem? Please describe.**

I found some issues when I migrated from springfox to springdoc. According to the [documentation](https://springdoc.org/#migrating-from-springfox), I just need to do the equivalent replacement of the annotation and I'll be able to get the migration done, but the fact slaps me.

1. @ParameterObject must be added when using objects as query parameters.
2. no solution about using object when submitting form data. Must use multiple property with @RequestPart
3. no solution about showing field description with @Schema in object when I use object as query param. Must use @Paramter with field.

**Describe the solution you'd like**

I reimplemented MethodParameter and org.springdoc.core.AbstractRequestService to solve the problem mentioned in the first part. However, this method is very troublesome, 

**Describe alternatives you've considered**

the access level of the code causes me to copy a lot of unnecessary code. I hope to change the access level of the code, so that I can accomplish my goal by inheriting and rewriting a few core methods.

**Additional context**

springfox
repo: https://github.com/NaccOll/springdoc-issue-demo/tree/springfox

![springfox](https://user-images.githubusercontent.com/26819913/179477310-7ee3a924-a075-4b12-ae67-f6f740d378ac.png)

springdoc
repo: https://github.com/NaccOll/springdoc-issue-demo/tree/springdoc

![springdoc](https://user-images.githubusercontent.com/26819913/179477344-c2317b41-4dd4-4fe9-bee7-7f4a0754c835.png)

springdoc-fix
repo: repo: https://github.com/NaccOll/springdoc-issue-demo/tree/springdoc-fix

![springdoc-fix](https://user-images.githubusercontent.com/26819913/179478042-597a5256-bd3e-4027-8135-74107c2522fd.png)

springdoc-custom
repo: repo: https://github.com/NaccOll/springdoc-issue-demo/tree/springdoc-custom

![springdoc-custom](https://user-images.githubusercontent.com/26819913/179477381-dd85f8ed-9b20-402d-9ce6-ddec1508128f.png)

bnasslahsen:
@NaccOll, 

Thank you for your efforts. 
Are you willing to propose a PR ?

NaccOll:
> @NaccOll,
> 
> Thank you for your efforts. Are you willing to propose a PR ?

ok

bnasslahsen:
@NaccOll,

The following test shows that with the latest SNAPSHOT, you will able to reach the same result as in springdoc-custom.

[...]

PR #1744, opened 2022-07-18
feat: springfox compatible

NaccOll:
change code access level and improve springfox compatible by config in default behavior.

[relation issue](https://github.com/springdoc/springdoc-openapi/issues/1743)

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/AbstractRequestService.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/AbstractRequestService.java
@@ -49,7 +49,12 @@ import javax.validation.constraints.Size;
 
 import com.fasterxml.jackson.annotation.JsonView;
 import io.swagger.v3.core.util.PrimitiveType;
+import io.swagger.v3.oas.annotations.enums.Explode;
 import io.swagger.v3.oas.annotations.enums.ParameterIn;
+import io.swagger.v3.oas.annotations.enums.ParameterStyle;
+import io.swagger.v3.oas.annotations.extensions.Extension;
+import io.swagger.v3.oas.annotations.media.ArraySchema;
+import io.swagger.v3.oas.annotations.media.ExampleObject;
 import io.swagger.v3.oas.models.Components;
 import io.swagger.v3.oas.models.OpenAPI;
 import io.swagger.v3.oas.models.Operation;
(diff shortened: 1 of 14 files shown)


NaccOll:
@bespaltovyj 
I  change access level and imrpove springfox compatible in this pr. But will this be bad, because I don't know the design of this project so well that the configuration I submit does not conform to the specification. I'm wondering if I should just keep the part that modifies the access level of the code so it can be merged faster.

bespaltovyj:
@NaccOll sorry, but I am not involved in this project so deep

NaccOll:
@bespaltovyj Sorry, I typed the wrong username due to autocompletion

NaccOll:
@bnasslahsen 
I change access level and imrpove springfox compatible in this pr. But will this be bad, because I don't know the design of this project so well that the configuration I submit does not conform to the specification. I'm wondering if I should just keep the part that modifies the access level of the code so it can be merged faster.

bnasslahsen:
@NaccOll, 

No problem for changing the access level if there a right justification.
Could you please add the related tests to your PR so we can understand why these changes are necessary ?

NaccOll:
@bnasslahsen 
I think we need to discuss the issue before adding test cases.  As I said in the issue, springdoc's compatibility with springfox is not as good as the documentation says.  More importantly, springdoc does not provide an out-of-the-box behavior consistent with spring mvc. For example, now springdoc treats query parameter objects as json or does not set parameters to form data when specifying api to accept form data.  These default behaviors will cause developers to spend more time writing swagger annotations by themselves.  In this pr, I did two things, one was to modify the access level of the code, and the other was to configure the three problems I said in the issue.  When writing test cases for code access level, do I need to roll back the solution to the problem mentioned in the issue?  It's embarrassing that modifying the code access level is to fix the default behavior bug.  However, I also solved the latter in pr.

bnasslahsen:
@NaccOll,

I have taken in consideration your proposal in this commit https://github.com/springdoc/springdoc-openapi/commit/fb4247868a17502e9c8e29140fe8c0b4efd41364
I didn't add the visibility change of methods and the properties as people can manage achieving it without these.

Thank again for your efforts to make the project adoption easier.

Do not hesistate, to propose additionnal to complete my actual commit. Just make sure the PRs do not mix a lot of changes.

PR #1749, opened 2022-07-21
Update server out of cache

### LINKED ISSUE #1748: Generated server url computation not cleared
Similar to #1712 , server url automatically generated mixed with customizer that modifies automatically generated server list, when used with both direct and reversed proxy (say with different context time for the same runtime), we end with a *first asked wins* as, at first call, open api is built and

* server url gets [computed](https://github.com/springdoc/springdoc-openapi/blob/5bd85e15ec0fca3a5b6e0a20f025e0cb943d6d53/springdoc-openapi-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java#L333)
* server list gets [copied (strange to use json serialisation)](https://github.com/springdoc/springdoc-openapi/blob/5bd85e15ec0fca3a5b6e0a20f025e0cb943d6d53/springdoc-openapi-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java#L342)
* customizer are [called](https://github.com/springdoc/springdoc-openapi/blob/5bd85e15ec0fca3a5b6e0a20f025e0cb943d6d53/springdoc-openapi-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java#L350)
* As server list is modified by customizer, it is considered that server are provided. 

The last assumption is wrong as if it was not provided at first.

In the end we ends with a customized generated server put in cache.

antechrestos:
Update servers out of cache and do not set server as provided when they are modified by customizer. This way we can still have a generic customized server address

Close #1748


antechrestos:
@bnasslahsen I spotted an attribute named `customiser` instead of `customizer` in `OpenAPIService`. I did not renamed it as it would have break some existing api call for api client; yet I can do it in another PR if you are ok.

bnasslahsen:
@antechrestos,

That's OK!

PR #1751, opened 2022-07-22
Fix build error due to invalid version of tests module

JohnNiang:
#### What kind of this PR

/kind bug

#### What this PR does / Why we need it

1. Make `springdoc-openapi-tests` as a submodule. After doing this, the versions in submodules will be automatically updated when the next release is made. Please refer to https://github.com/springdoc/springdoc-openapi/commit/9528d8f7f52c22e6c10aa43dd611d41c7ed026d6.
2. Align versions of test modules with the main project to prevent build failures

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,6 +79,7 @@
 		<module>springdoc-openapi-starter-webflux-api</module>
 		<module>springdoc-openapi-starter-webmvc-ui</module>
 		<module>springdoc-openapi-starter-webflux-ui</module>
+		<module>springdoc-openapi-tests</module>
 	</modules>
 
 	<properties>
--- a/springdoc-openapi-tests/pom.xml
+++ b/springdoc-openapi-tests/pom.xml
@@ -4,7 +4,7 @@
 	<parent>
 		<artifactId>springdoc-openapi</artifactId>
 		<groupId>org.springdoc</groupId>
-		<version>2.0.0-M3-SNAPSHOT</version>
+		<version>2.0.0-M4-SNAPSHOT</version>
 	</parent>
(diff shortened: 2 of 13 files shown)


bnasslahsen:
@JohnNiang, 

Totally agree 76d3dd68. 
Thank you, just saw your PR.

pom.xml
```diff
@@ -79,6 +79,7 @@
 		<module>springdoc-openapi-starter-webflux-api</module>
 		<module>springdoc-openapi-starter-webmvc-ui</module>
 		<module>springdoc-openapi-starter-webflux-ui</module>
+		<module>springdoc-openapi-tests</module>
```
JohnNiang, line 82: Hi, @bnasslahsen , please note the new module `springdoc-openapi-tests` here. After adding it, https://github.com/springdoc/springdoc-openapi/commit/3eddb9d88dfde16be3224ab49139b09d2738cf46 and https://github.com/springdoc/springdoc-openapi/commit/72265643923890b312eb05cf42233a5aadc02267 will update the version automatically instead of updating manually.

If we care about the deployment for the module `springdoc-openapi-tests`, don't forget to configure the plugin `maven-deploy-plugin` at `springdoc-openapi-tests/pom.xml`:

```xml
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-deploy-plugin</artifactId>
  <configuration>
    <skip>true</skip>
  </configuration>
</plugin>
```

bnasslahsen: @JohnNiang, 

Thank you for your hint. For `springdoc-openapi-tests` it won't be necessary to keep deploying this module.

PR #1754, opened 2022-07-23
fix: to adjust default value and description of the page parameter when enabled `spring.data.web.pageable.one-indexed-parameters` property.

marcospds:
When the property `spring.data.web.pageable.one-indexed-parameters` is enabled, the page starts at 1 and not at 0, so the description and the default value are incoherent.

To simulate the behavior this project can be used: https://github.com/marcospds/springdoc-one-indexed-parameters


bnasslahsen:
Thank you @marcospds for your contribution to the project!

PR #1755, opened 2022-07-25
fix: to adjust name of the parameters page and size when set `spring.data.web.pageable.prefix` property.

marcospds:
When the property `spring.data.web.pageable.prefix` is seted, the name of the page and size parameters are incorrect.

To simulate the behavior this project can be used: https://github.com/marcospds/springdoc-prefix-parameter

Before:
antes

After:
depois


bnasslahsen:
Thank you @marcospds for your contribution to the project!

PR #1855, opened 2022-09-22
Fixed a bug that duplicate field were get for record classes

uc4w6c:
## Describe the bug
When the argument of the MethodParameterPojoExtractor.extractFrom method is a record class, duplicate fields gets.

1. when the argument is a class (expected)

```java
// The ClassObject class has email, firstName, lastName fields and getters 
Stream<MethodParameter> actual = MethodParameterPojoExtractor.extractFrom(ClassObject.class);
actual.forEach(method -> System.out.println(method.getMethod().getName()));

// Below is the output result.
// getEmail
// getFirstName
// getLastName
```

2. when the argument is a record (unexpected)

```java
record RecordObject(String email, String firstName, String lastName) {}
Stream<MethodParameter> actual = MethodParameterPojoExtractor.extractFrom(RecordObject.class);
actual.forEach(method -> System.out.println(method.getMethod().getName()));

// Below is the output result.
// email
// firstName
// lastName
// email
// firstName
// lastName
// email
// firstName
// lastName
```

## Fixes
Added filter as well as Class.


bnasslahsen:
Thank you @uc4w6c for your contribution to the project.

Issue #96, opened 2019-09-26
/v3/api-docs inconsistence of operationId


SleepyDeb:
Using the latest version (1.1.42) the api returns "flip-flopping" the operation id with "_1" as a suffix

First Time:
> "operationId":"getUsers"

After:
> "operationId":"getUsers_1"

Then:
> "operationId":"getUsers"

and again

thanks in advance

maciejkujawski:
It affects also other fields... With my schema openAPi.yaml at first generation is correct, when I refresh swagger or openApi url new file is wrong. Eg. First invocation:

```
     operationId: getFile
      parameters:
      - name: canonicalPath
        in: query
        description: xxx
        required: true
        schema:
          type: string
      responses:
        200:
          description: File resource
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/File'
        400:
          description: Wrong/malformed request parameters
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Error'
```

Second refresh:

```
     operationId: getFile_1
      parameters:
      - name: canonicalPath
        in: query
        description: xxx
        required: true
        schema:
          type: string
      responses:
        200:
          description: File resource
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/File'
        400:
          description: Wrong/malformed request parameters
          content:
            application/json:
              schema:
                oneOf:
                - $ref: '#/components/schemas/File'
                - $ref: '#/components/schemas/Error'
```

On version 1.1.39 only changing operation id is affected. On newest 1.1.43 both error schema and operationId are wrong. Moreover the order of paths is different when I generate openApi.yaml/json with 1.1.39 vs 1.1.43

springdoc:
Hi maciejkujawski,

 The id operaration, generated dynamically should not have an impact on other  fields.
 Can you please add sample code to reproduce this behaviour.

maciejkujawski:
You can reproduce it with:

```
  @Operation(description = "Download file")
  @ApiResponse(responseCode = "200", description = "File resource", content = @Content(schema = @Schema(implementation = java.io.File.class)))
  @ApiResponse(responseCode = "400", description = "Wrong request", content = @Content(schema = @Schema(implementation = java.lang.Error.class)))
  @ApiResponse(responseCode = "500", description = "Unexpected error", content = @Content(schema = @Schema(implementation = java.lang.Error.class)))
  @GetMapping(value = "/file",
[...]

PR #1878, opened 2022-10-07
Sort request methods

yotam2010:
There is a bug with a random integer suffix for generated operation id when there is `@RequestMapping` without a specific method.

Similar to another issue that was solved already: https://github.com/springdoc/springdoc-openapi/issues/96

But the solution introduced for the closed issue didn't help in this case.

Added a test to show it is solved. You can reproduce the issue by reverting from TreeSet to HashSet and running the test multiple times (you will see you get different operation IDs randomly).

The reason is, HashSet does not guarantee the same order, so in different runs, it may produce different orders and therefor different operation ids.

It is my first open-source contribution so please let me know if I missed anything in the process.

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java
@@ -42,6 +42,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.TreeSet;
 import java.util.concurrent.CopyOnWriteArraySet;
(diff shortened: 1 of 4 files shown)


yotam2010:
Output example of the bug (running the test I added twice without TreeSet):
[Expected output](https://jsoneditoronline.org/#left=cloud.38c09d4deed341d289456844bc7fa326)
[Actual output](https://jsoneditoronline.org/#left=cloud.e86b145064fb4d7cac20b54da1503b6a)

yotam2010:
Hey @bnasslahsen, I saw you are active here, can you take a look at my PR?

bnasslahsen:
@yotam2010,

Thank you for your contribution to the project1

PR #1907, opened 2022-10-24
Configuration compiler : JDK 17

xuxiaowei-com-cn:


xuxiaowei-com-cn:
The submitted content can be configured as `profile` in `setting. xml` by default.
More convenient packaging.

bnasslahsen:
@xuxiaowei-com-cn,

This is inherited from spring-boot parent pom.

PR #1959, opened 2022-11-26
Replaced deprecated org.springframework.core.LocalVariableTableParameterNameDiscoverer with org.springframework.core.DefaultParameterNameDiscoverer

krupt:


bnasslahsen:
@krupt,

I have fixed your PR:
LocalVariableTableParameterNameDiscoverer should have been replaced by StandardReflectionParameterNameDiscoverer.

krupt:
But why? StandardReflectionParameterNameDiscoverer can't work for every library. DefaultParameterNameDiscoverer is better, cause it's uses all posible options. First Kotlin, if exists, then StandardReflectionParameterNameDiscoverer for classes that were compiled with Java 8+ with -parameters option and then LocalVariableTableParameterNameDiscoverer for all other classes. That's how it works in Spring Framework

bnasslahsen:
You were breaking tests ...

sascha-sphw:
In which version will this be included?

bnasslahsen:
Read https://github.com/springdoc/springdoc-openapi/releases

PR #1969, opened 2022-11-29
Prevents premature initialisation of factory-beans. Fixes #1965

### LINKED ISSUE #1965: Change of behaviour in the beans loading order
**Describe the bug**

Hello

I wanted to tell you about the problem I am having. The problem is unusual and will not happen easily.

The _BeanFactoryPostProcessors_ [SpringdocBeanFactoryConfigurer ](https://github.com/springdoc/springdoc-openapi/blob/b274a0e517cc73c3f7a809f48b9b3771cf840308/springdoc-openapi-common/src/main/java/org/springdoc/core/SpringdocActuatorBeanFactoryConfigurer.java#L47) and [SpringdocActuatorBeanFactoryConfigurer ](https://github.com/springdoc/springdoc-openapi/blob/b274a0e517cc73c3f7a809f48b9b3771cf840308/springdoc-openapi-common/src/main/java/org/springdoc/core/SpringdocActuatorBeanFactoryConfigurer.java#L47) are causing a change in the loading order in an application I have with **springdoc** and **spring-cloud-stream** when there is a _pollable_ and it is causing an error.

_ConfigurableListableBeanFactory#getBeanNamesForType(Class_) method is causing factorybeans to be initialised. 

https://github.com/springdoc/springdoc-openapi/blob/b274a0e517cc73c3f7a809f48b9b3771cf840308/springdoc-openapi-common/src/main/java/org/springdoc/core/SpringdocBeanFactoryConfigurer.java#L60-L64

Internally this call is being made:

```java
	@Override
	public String[] getBeanNamesForType(@Nullable Class<?> type) {
		return getBeanNamesForType(type, true, true);
	}
```

And according to the javadoc of [ListableBeanFactory](https://github.com/spring-projects/spring-framework/blob/164141a63e93c5ad326f41441fc475e5ebaf6cdb/spring-beans/src/main/java/org/springframework/beans/factory/ListableBeanFactory.java#L249), the third parameter can cause dangerous behaviour.

> allowEagerInit – whether to initialize lazy-init singletons and objects created by FactoryBeans (or by factory methods with a "factory-bean" reference) for the type check. Note that FactoryBeans need to be eagerly initialized to determine their type: So be aware that passing in "true" for this flag will initialize FactoryBeans and "factory-bean" references.

In my case,  [BindableFunctionProxyFactory](https://github.com/spring-cloud/spring-cloud-stream/blob/v3.2.6/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/BindableFunctionProxyFactory.java) (implements FactoryBean) is being initialised early and internally doing 

```java
	protected void populateBindingTargetFactories(BeanFactory beanFactory) {
		this.bindingTargetFactories = ((ListableBeanFactory) beanFactory).getBeansOfType(BindingTargetFactory.class);
	}
```

. But as it is initialised early, it will not load all the beans it should. That is, the FactoryBean is created without guaranteeing that all the beans have already been created.

**To Reproduce**

- What version of spring-boot you are using? 2.7.6
- What modules and versions of springdoc-openapi are you using? It happens with both springdoc-webmvc and springdoc-webflux.
- What version of springdoc you are using? 1.5.x, 1.6.x and 2.0.0
- Occurs with _springdoc.show-actuator=true_ and _managemente.port_ different.

**Solution**

This problem can easily be solved in the following way 

In **SpringdocBeanFactoryConfigurer** class, get beans by type with allowEagerInit flag set to false

```java
  /**
   * Init bean factory post processor.
   *
   * @param beanFactory the bean factory
   */
  public static void initBeanFactoryPostProcessor(final ConfigurableListableBeanFactory beanFactory) {
    for (final String beanName : beanFactory.getBeanNamesForType(OpenAPIService.class, true, false)) {
      beanFactory.getBeanDefinition(beanName).setScope(SCOPE_PROTOTYPE);
    }
    for (final String beanName : beanFactory.getBeanNamesForType(OpenAPI.class, true, false)) {
      beanFactory.getBeanDefinition(beanName).setScope(SCOPE_PROTOTYPE);
    }
  }

```

In fact, I can make a contribution if you like.

jorgerod:
Closes #1965 

NOTE 1: I do the PR against main as the contribution guide says. Do you do it against springdoc?
NOTE 2: I have not included unit tests as the integration tests currently available cover this feature.
NOTE 3: Test [test.org.springdoc.api.v30.app114.SpringDocApp114Test.testApp](http://129.159.254.115:8686/job/springdoc-openapi-PR/org.springdoc$springdoc-openapi-webmvc-core/204/testReport/junit/test.org.springdoc.api.v30.app114/SpringDocApp114Test/testApp/) was already failing beforehand

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/SpringdocBeanFactoryConfigurer.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/SpringdocBeanFactoryConfigurer.java
@@ -58,9 +58,9 @@ public class SpringdocBeanFactoryConfigurer implements EnvironmentAware, BeanFac
 	 * @param beanFactory the bean factory
 	 */
 	public static void initBeanFactoryPostProcessor(ConfigurableListableBeanFactory beanFactory) {
-		for (String beanName : beanFactory.getBeanNamesForType(OpenAPIService.class))
+		for (String beanName : beanFactory.getBeanNamesForType(OpenAPIService.class, true, false))
 			beanFactory.getBeanDefinition(beanName).setScope(SCOPE_PROTOTYPE);
-		for (String beanName : beanFactory.getBeanNamesForType(OpenAPI.class))
+		for (String beanName : beanFactory.getBeanNamesForType(OpenAPI.class, true, false))
 			beanFactory.getBeanDefinition(beanName).setScope(SCOPE_PROTOTYPE);
 	}
 


bnasslahsen:
Thank you @jorgerod for your PR
It's ok. Tests are fixed as well.

PR #2052, opened 2023-01-24
Fix IntelliJ instructions

westse:
- Should reference pom.xml, not build.gradle (which doesn't exist).


bnasslahsen:
@westse,

PR #2053, opened 2023-01-24
Support maven wrapper

westse:
- Better build reproducibility: ensures Maven version is defined by the project itself.
- Convenience: wrapper takes care of downloading (the correct version of) Maven for you.
- See https://maven.apache.org/wrapper, similar to [Gradlew Wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html)
- Used by a number of spring projects (e.g. https://github.com/spring-projects/spring-plugin)


bnasslahsen:
@westse,

Thank you for your PR.
We do not want ship any jar files in the source repo.

Thank you for your understanding.

Issue #2054, opened 2023-01-25
Add copyright and license information to Jar.


tobi5775:
For a customer project an OpenSource report has to be generated. Therefore a scanner was created to extract the copyright and licence information from our dependencies. This scanner can extract all informaton from the Jar File. Therefore it would be great to have the copyright and license information included in the Jar File.

PR #2055, opened 2023-01-25
Issue #2054: Add copyright and license information to Jar.

tobi5775:
For a customer project an OpenSource report has to be generated. Therefore a scanner was created to extract the copyright and licence information from our dependencies. This scanner can extract all informaton from the Jar File. Therefore it would be great to have the copyright and license information included in the Jar File.


bnasslahsen:
@tobi5775.

PR #2067, opened 2023-02-09
Avoid `example: null` when sorted

westse:
- Seen when `springdoc.writer-with-order-by-keys` is enabled, but not when disabled.
- Avoids unnecessary entry in api specification.

While this fix seems good, I wonder what the original intent was of using `JsonInclude.Include.CUSTOM`...


bnasslahsen:
@westse,

Can you please add the related tests that shows why you need this change ?

Thank you so much for your contribution

xin-li-ui (approved):

bnasslahsen:
@westse,

This PR will be closed, as no test has been provided.

Issue #2101, opened 2023-03-02
@Deprecated is ignored when enumsAsRef==true / $refs can't be deprecated


daniel-frak:
**Describe the bug**

When `enumsAsRef` is set to `true`, the `@Deprecated` annotation on fields of type enum is ignored.

**To Reproduce**

- Using Spring Boot 3.0.3
- Using `springdoc-openapi-starter-common` and `springdoc-openapi-starter-webmvc-ui` version `2.0.2`

Given an example endpoint:
```java
import org.springframework.context.annotation.Configuration;
import org.springframework.web.bind.annotation.GetMapping;
import org.springframework.web.bind.annotation.RestController;

@RestController
public class TestController {

    @GetMapping
    public MyClass testEndpoint() {
        return null;
    }

    public enum TestEnum {

        VAL1, VAL2
    }

    @Configuration
    public static class OpenApiConfig {

        static {
            // Generate enums on frontend
            io.swagger.v3.core.jackson.ModelResolver.enumsAsRef = true;
        }
    }

    public record MyClass(
            TestEnum field1,
            @Deprecated TestEnum field2
    ) {
    }
}
```

The produced OpenAPI schema is:
```yaml
openapi: 3.0.1
info:
  title: OpenAPI definition
  version: v0
servers:
- url: http://localhost:8080
  description: Generated server url
paths:
  /:
    get:
      tags:
      - test-controller
      operationId: testEndpoint
      responses:
        "200":
          description: OK
          content:
            '*/*':
              schema:
                $ref: '#/components/schemas/MyClass'
components:
  schemas:
    MyClass:
      type: object
      properties:
        field1:
          $ref: '#/components/schemas/TestEnum'
        field2:
          $ref: '#/components/schemas/TestEnum'
    TestEnum:
      type: string
      enum:
      - VAL1
      - VAL2
```

That is, the `@Deprecated` annotation gets ignored.

Removing the line io.swagger.v3.core.jackson.ModelResolver.enumsAsRef = true; produces correct output (the deprecation gets applied to the field) but, of course, the enum is not separately documented:
```yaml
openapi: 3.0.1
info:
  title: OpenAPI definition
  version: v0
servers:
- url: http://localhost:8080
  description: Generated server url
paths:
  /:
    get:
      tags:
      - test-controller
      operationId: testEndpoint
      responses:
        "200":
          description: OK
          content:
            '*/*':
              schema:
                $ref: '#/components/schemas/MyClass'
components:
[...]

PR #2116, opened 2023-03-10
#2101 Fix @Deprecated annotation being ignored on fields using $ref (i.e. fields referencing classes)

daniel-frak:


bnasslahsen:
Not approved

PR #2183, opened 2023-03-30
fixed DefaultFlatParamObject to work with annotated parameters. Fixes…

### LINKED ISSUE #2181: default-flat-param-object does not take effect
If you ❤️ this project consider becoming a [sponsor](https://github.com/sponsors/springdoc).
If you already use the `springdoc-openapi`, please do not forget adding a ***github star*** to this repository.

**Describe the bug**

I upgraded the springdoc-openapi dependencies from 1.6.9 to 1.6.15, and then I tried to use the new feature `default-flat-pam-object`, but it didn't work after the configuration was complete.

**To Reproduce**

Spring Boot 2.7.10

springdoc-openapi 1.6.15

I had to make extensive use of the @ParameterObject annotation before, and I hope that with this configuration I can get rid of this hassle.

```yml
springdoc:
  swagger-ui:
    path: /swagger-ui.html
    tags-sorter: alpha
    operations-sorter: alpha
    show-extensions: true
  api-docs:
    enabled: ${springdoc.swagger-ui.enabled}
    path: /v3/api-docs
  group-configs:
    - group: 'default'
      paths-to-match: '/**'
      packages-to-scan: top.charles7c.cnadmin.webapi.controller
  default-flat-param-object: true
```
[https://github.com/Charles7c/continew-admin](https://github.com/Charles7c/continew-admin)

**Screenshots**
image

uc4w6c:
fix #2181

Since there is no test code in the repository from before, I don't know why it is doing annotation checks.
However, since the target parameter can be determined by `AbstractRequestService.isRequestTypeToIgnore`, I would like to remove the annotation retention confirmation process.

Related PR: https://github.com/springdoc/springdoc-openapi/pull/1805/files

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/DelegatingMethodParameter.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/DelegatingMethodParameter.java
@@ -125,21 +125,12 @@ public class DelegatingMethodParameter extends MethodParameter {
 					explodedParameters.add(methodParameter);
 				});
 			}
-			else if (defaultFlatParamObject) {
-				boolean isSimpleType = MethodParameterPojoExtractor.isSimpleType(paramClass);
-				boolean hasAnnotation = p.hasParameterAnnotations();
-				boolean shouldFlat = !isSimpleType && !hasAnnotation;
-				if (shouldFlat && !AbstractRequestService.isRequestTypeToIgnore(paramClass)) {
-					MethodParameterPojoExtractor.extractFrom(paramClass).forEach(methodParameter -> {
-						optionalDelegatingMethodParameterCustomizer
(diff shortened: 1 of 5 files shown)


bnasslahsen:
@uc4w6c,

NaccOll:
@Charles7c @uc4w6c 
DefaultFlatParamObject is what I added to reduce the usage of ParameterObject. Stopping automatic flattening of objects when annotations are found is itself an expected behavior, because when I annotate an object, I don't want him to be flattened (like @RequestBody, @RequestPart). If you want to add annotations to keep things flat, you should add configuration to ignore specified annotations instead of removing this code directly. Because of this pr, there are problems with all methods using @RequestBody and @RequestPart at present, and the document cannot be displayed correctly

PR #2194, opened 2023-04-04
#2193 Fix Swagger UI with provided spec

### LINKED ISSUE #2193: Swagger UI with provided spec broken in 1.7.0
**Describe the bug**

Using Springdoc + Swagger UI with a provided spec as described [here](https://springdoc.org/#what-is-a-proper-way-to-set-up-swagger-ui-to-use-provided-spec-yml) seems to be broken in version `1.7.0`. Attempting to navigate to `swagger-ui.html` yields `404 Not Found`.

**To Reproduce**

`2.7.10`
  - `springdoc-openapi-ui`
  - `springdoc-openapi-security`

**Expected behavior**

The description found [here](https://springdoc.org/#what-is-a-proper-way-to-set-up-swagger-ui-to-use-provided-spec-yml) should either be updated to reflect additional steps or should work again.

**Additional context**

The problem appears to be the newly added `org.springdoc.core.SpringDocUIConfiguration`, which is only loaded when property `springdoc.api-docs.enabled` is set to `true`. With the aforementioned description to use a provided spec, said property needs to be set to `false`.

Maybe it is possible to simply remove `@ConditionalOnExpression("${springdoc.api-docs.enabled:true}")` from `SpringDocUIConfiguration`? As an alternative, the [minimal Beans configuration](https://springdoc.org/#what-is-a-proper-way-to-set-up-swagger-ui-to-use-provided-spec-yml) might be expanded to include the contents of `SpringDocUIConfiguration`.

tobi-laa:
Fixes #2193.

For cases where Swagger UI should be deactivated completely, `@ConditionalOnBean(SpringDocConfiguration.class)` should still be sufficient for excluding the config - at least that is my understanding :)

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/core/SpringDocUIConfiguration.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/core/SpringDocUIConfiguration.java
@@ -32,7 +32,6 @@ import org.apache.commons.lang3.StringUtils;
 
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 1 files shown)


bnasslahsen:
@tobi-laa,

PR #2198, opened 2023-04-07
Refactored a few design and implementation smells

vidhimaheriya:


bnasslahsen:
@vidhimaheriya,

Can you propose the same PR for the master branch as well ?

bnasslahsen:
depends on #2206
which is closed, for no activity since 3 weeks ago

PR #2213, opened 2023-04-21
Using both generated and configured specs stoped working in 1.6.5

bjornblomqvist:
Hi, prior to version 1.6.5 I could you use both a spec file and generated spec from annotations. This behavior is something I need to be able to upgrade to a later version. I am not sure if this feature that I have been using was even ment to work but it did and I would very much appreciate if we could bring it back.

You can find in this pull request a test and a fix for my problem. I think I put the test in the wrong place so I would appreciate if you could help me improve the test. Also as I don't know the code I think the fix might also be in the wrong place.

Best regards
Björn Blomqvist

### DIFF
--- a/springdoc-openapi-common/src/main/java/org/springdoc/ui/AbstractSwaggerWelcome.java
+++ b/springdoc-openapi-common/src/main/java/org/springdoc/ui/AbstractSwaggerWelcome.java
@@ -132,10 +132,11 @@ public abstract class AbstractSwaggerWelcome {
 			else
 				swaggerUiConfigParameters.addUrl(apiDocsUrl);
 			if (!CollectionUtils.isEmpty(swaggerUiConfig.getUrls())) {
-				swaggerUiConfigParameters.setUrls(swaggerUiConfig.cloneUrls());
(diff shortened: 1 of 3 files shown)


bjornblomqvist:
I cant see the result of the failing build as the requirement for accessing it is to give the build system access to my private repos. Witch I obviously will not do.


But I am guessing that my change is not causing the build failure as it works on 1.7.0 and the master branch already has build failures.

bjornblomqvist:
@bnasslahsen what do you think about this merge request?

bjornblomqvist:
The only change was to rebase on latest master

bjornblomqvist:
Last force push was to remove my employers email address.

bjornblomqvist:
@bnasslahsen have you had time to look at this yet?

bnasslahsen:
@bjornblomqvist,

Not yet. 
But it will be my next PR in the project to review.

bnasslahsen:
Thank you @bjornblomqvist for your PR.
It's now merged and also committed to 2.x branch as well.

PR #2240, opened 2023-05-24
upgrade spring-boot version

nfawcett:


bnasslahsen:
already updated to spring-boot 2.7.12

Issue #2235, opened 2023-05-19
Trailing slash causes 500 when loading /swagger-ui/


wtell400:
**Describe the bug**

Given
- Spring Boot 3.0.2 (using default for PathMatchConfigurer.setUseTrailingSlashMatch(boolean))
- springdoc-openapi 2.1.0

springdoc.api-docs.enabled=true
springdoc.api-docs.path=/api/docs
springdoc.swagger-ui.enabled=true
springdoc.swagger-ui.path=/api/docs/swagger-ui

When 
accessing swagger-ui using:
https://myhost/api/docs -> loads openapi json
https://myhost/api/docs/ -> 404 (expected)
https://myhost/api/docs/swagger-ui -> redirects to https://myhost/api/docs/swagger-ui/index.html, loads swagger-ui
https://myhost/api/docs/swagger-ui/ -> 500 (unexpected)

```
java.lang.StringIndexOutOfBoundsException: begin 11, end 10, length 10
	at java.base/java.lang.String.checkBoundsBeginEnd(String.java:4604)
	at java.base/java.lang.String.substring(String.java:2707)
	at java.base/java.lang.String.substring(String.java:2680)
	at org.springdoc.ui.AbstractSwaggerResourceResolver.path(AbstractSwaggerResourceResolver.java:71)
	at org.springdoc.ui.AbstractSwaggerResourceResolver.findWebJarResourcePath(AbstractSwaggerResourceResolver.java:42)
	at org.springdoc.webmvc.ui.SwaggerResourceResolver.resolveResource(SwaggerResourceResolver.java:33)
	at org.springframework.web.servlet.resource.DefaultResourceResolverChain.resolveResource(DefaultResourceResolverChain.java:75)
...<lots of calls unrelated to sprindoc>...
```

**To Reproduce**
See the webmvc demo site:

http://144.24.171.248:8081/ -> redirects to http://144.24.171.248:8081/swagger-ui/index.html, loads swagger-ui
http://144.24.171.248:8081/swagger-ui -> 500
http://144.24.171.248:8081/swagger-ui/ -> 500

**Expected behavior**

https://myhost/api/docs/ -> 404
https://myhost/api/docs/swagger-ui/ -> 404

**Additional context**
[...]

PR #2243, opened 2023-05-27
fix StringIndexOutOfBoundsException when path is same webjar. #2235

uc4w6c:
#2235


bnasslahsen:
@uc4w6c,

Thank you for your great involvement in the project!

PR #2280, opened 2023-06-26
fix loop when reponse inherits generic class #2271

marcospds:
#2271

The code was a recursive loop when the generics of resolvableTypes are the same resolvableTypes.

Before (2.0.9):
image

After (2.1.1-SNAPSHOT fixed):
image

App of to simulation:
https://github.com/marcospds/springdoc-inherits-generic-class


bnasslahsen:
@marcospds,

Issue #2282, opened 2023-06-27
default-flat-param-object doesn't work when using http body


NaccOll:
**Describe the bug**

Because of an [issue](https://github.com/springdoc/springdoc-openapi/issues/2181), a [pr](https://github.com/springdoc/springdoc-openapi/pull/2183) was proposed, but this led to a bigger problem. All annotations that use RequestBody, RequestPart and custom message body processing are rendered invalid on swagger-ui. Objects become flattened objects and eventually query parameters, which cannot be processed into json or other required message body formats.

**To Reproduce**

- What version of spring-boot you are using?
  Spring Boot 3.1.1
- What modules and versions of springdoc-openapi are you using?
  springdoc-openapi 2.1.0
- What is the actual and the expected result using OpenAPI Description (yml or json)?
  When I use default-flat-param-object, I want swagger-ui to behave in line with how spring web handles
- Provide with a sample code (HelloController) or Test that reproduces the problem
```yaml
springdoc:
  default-flat-param-object: true
```

**Expected behavior**

I want springdoc to automatically flatten my complex query param objects when I use default-flat-param-object, but it shouldn't do anything for the json objects I need to accept as well as form objects. But now it turns my json object into a query parameter, and my swagger-ui is not working properly.

In fact, the emergence of default-flat-param-object is to solve the problem of repeatedly adding ParameterObject. But more importantly, the core problem is that it is close to springweb's default behavior. The simple way is of course to only process RequestBody and RequestPart, but I know that many developers will customize annotations to process requests. In order to be compatible with this part of developers, when using annotations for parameters, ignore default-flat-param-object

**Screenshots**
![image](https://github.com/springdoc/springdoc-openapi/assets/26819913/1589c312-5862-499a-ad15-1f11319684e8)

```java
@PreAuthorize("hasPermission(0L,'GLOBAL','cms/column:add')")
@Operation(summary = "创建栏目")
@PostMapping
public CmsColumnDto create(@RequestBody CmsColumnCreateCommand command) {
     ...
}
```
`command` is a json object but converted to multiple query param.

PR #2283, opened 2023-06-27
fix: https://github.com/springdoc/springdoc-openapi/issues/2282

NaccOll:


bnasslahsen:
@NaccOll,

Can you add the Sample HelloController related to your PR, with an example of the impacted code?
If you can rebase your PR on the main branch, it will be great - so it can be merged directly.

Issue #2299, opened 2023-07-11
Performance Issue: springdoc-openapi


Avanishkumar28:
I'm having performance issue with my project. I have 100+ endpoint server took approximately ~15 mins (some times even more).
Although I have set property `springdoc.pre-loading-enabled: true` but when I hit first request from the browser, server again initialising `o.springdoc.api.AbstractOpenApiResource`
So in my server logs I can see
`2023-07-10T09:52:31.103Z  INFO 22964 --- [pool-12-thread-1] o.springdoc.api.AbstractOpenApiResource  : Init duration for springdoc-openapi is: 853743 ms`

and again when I hit first request from browser-

`2023-07-10T10:10:14.392Z  INFO 22964 --- [io-13080-exec-1] o.springdoc.api.AbstractOpenApiResource  : Init duration for springdoc-openapi is: 828973 ms`

No Idea why it is re-initialising.

System details
Windows 11, 16 GB RAM, Processor- AMD EPYC 7763 64-Core

PR #2309, opened 2023-07-15
feat: Added function to preload by specifying locale

uc4w6c:
fix https://github.com/springdoc/springdoc-openapi/issues/2299.

When `springdoc.pre-loading-enabled: true` is set, the default locale is specified and openapi is generated, but if another locale is specified from the browser, openapi needs to be generated again .

Improve performance by specifying the locale you want to preload and generating openapi in advance.

Additional explanation
For example, in Japan, the default locale is `ja_JP`, but the ACCEPT_LANGUAGE specified by the browser is `ja`, so it is not preloaded.


uc4w6c:
If the PR is fine, I'll create a doc.

bnasslahsen:
Looks fine thank you!

Issue #2256, opened 2023-06-07
SpringDoc bean naming conflict error with GraphQL Spring boot starter


ghoshrahul:
In a spring boot project which is using version 3.0.x and springdoc-openapi 2.x, there is a bean conflict for `ObjectMapperProvider`.

- What version of spring-boot you are using? **3.0.7**
- What modules and versions of springdoc-openapi are you using? **2.1.0**

**Expected behavior**

- Application with spring boot graphql starter and spring doc should have no conflicts and be running.

**Additional context**
Refer this [link](https://github.com/graphql-java-kickstart/graphql-spring-boot/issues/909)

uc4w6c:
@ghoshrahul 
I changed the bean name. It will be resolved in the next version.

PR #2310, opened 2023-07-16
fix: change bean name of objectMapperProvider

uc4w6c:
issue #2256


bnasslahsen:
Thank you @uc4w6c for your help!

Issue #130, opened 2019-10-28
Multiple models with the same class name will be parsed into the same


tomgao365:
`cc.tomgao.demo.springdoc.crm.model.User` and `cc.tomgao.demo.springdoc.sys.model.User` 

![image](https://user-images.githubusercontent.com/13798931/67645590-96967780-f964-11e9-8782-0ccd084a6213.png)

[demo](https://github.com/my-alpha/springdoc-openapi-demo)

springdoc:
Hi, 

 By default, Both of your models will use the same schema name, which is the class name. 
 You can just provide a different schema name, for each class in this case:
 
```java
package cc.tomgao.demo.springdoc.sys.model.User;
@Schema(name = "User1")
public class User
```
and:
```java
package cc.tomgao.demo.springdoc.crm.model.User;
@Schema(name = "User2")
public class User
```

olaisolheim:
The solution outlined in this solution is not really a solution, but a workaround, which requires access to alter the classes, which is not always the case.

Spring Doc assumes classes with the same name will also have the same schema/contract, which is obviously not a great assumption to begin with.

Is there another way to have Spring Doc differentiate between the two?

PR #2319, opened 2023-07-24
Same model name converter

dsankouski:
add new converter to rename model name duplications #130

Finds model classes with same name but different package,
and renames model in specification.
Ignore SameModelNameConverter in test 114, because AdditionalModelsConverter
used to pick the right class from 2 classes with the same name.

add unit test for SameModelConverter #130


bnasslahsen:
@dsankouski,

For this case, why don't you use: `springdoc.use-fqn=true` ?

dsankouski:
@bnasslahsen The spec is used to generate java client for automated testing. In case of fqn model classname would be really long.

bnasslahsen:
@dsankouski,

Thank you for your PR.
You can use your `ModelConverter` for this particular usage.

We prefer to rely on the swagger-core standard option: `use-fqn`

Issue #2312, opened 2023-07-17
Spec for @ParameterObject disappears if building native-images


mykhailokulakov:
**Describe the bug**

Hey
I'm using spring-boot 3.1.1 and springdoc-openapi 2.1.0
When annotating the Pageable interface with @ParameterObject, everything works smoothly for a traditional build, with parameters being generated as expected:

```
"parameters": [
  {
	"name": "productId",
	...
  },
  {
	"name": "userId",
	...
  },
  {
	"name": "page",
	"in": "query",
	"description": "Zero-based page index (0..N)",
	"required": false,
	"schema": {
	  "minimum": 0,
	  "type": "integer",
	  "default": 0
	}
  },
  {
	"name": "size",
	"in": "query",
	"description": "The size of the page to be returned",
	"required": false,
	"schema": {
	  "minimum": 1,
	  "type": "integer",
	  "default": 20
	}
  },
  {
	"name": "sort",
	"in": "query",
	"description": "Sorting criteria in the format: property,(asc|desc). Default sort order is ascending. Multiple sort criteria are supported.",
	"required": false,
	"schema": {
	  "type": "array",
	  "items": {
		"type": "string"
	  }
	}
  }
],
```

But if running a native image, this part is missing from the generated spec:
```
"parameters": [
  {
    "name": "productId",
    ...
  },
  {
    "name": "userId",
    ...
  }
],
```

**To Reproduce**

spring-boot-starter-parent 3.1.1
springdoc-openapi-starter-webmvc-ui 2.1.0

```
@RestController
@RequestMapping("/tickets")
@RequiredArgsConstructor
@FieldDefaults(level = AccessLevel.PRIVATE, makeFinal = true)
@Tag(name = "Tickets API")
public class TicketController  {
    TicketService service;

    @GetMapping
    @Operation(summary = "Search tickets")
    @PageableAsQueryParam
    ResponsePage<Ticket> search(
            @RequestParam(value = "productId", required = false) UUID productId,
            @RequestParam(value = "userId", required = false) UUID userId,
            @Parameter(hidden = true) Pageable page) {
        return service.search(productId, userId, page);
    }
}
```

`mvn -Pnative spring-boot:build-image`

**Expected behavior**

It should work on a native-image

**Screenshots**

regular build:

![Screenshot 2023-07-18 002054](https://github.com/springdoc/springdoc-openapi/assets/3352223/b54917a4-7c39-4b2e-aabc-8350db1aea26)

native:
[...]

PR #2334, opened 2023-08-06
fix: Add RuntimeHints to work with ParameterObject

uc4w6c:
fix: #2312

code that is reflecting
https://github.com/springdoc/springdoc-openapi/blob/main/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/extractor/DelegatingMethodParameter.java##L171-L173


bnasslahsen:
@uc4w6c,

The problem is indeed reflection.
I didn't succeed to use your fix.
At the end i added few Hints for other use cases as well.

uc4w6c:
@bnasslahsen 
Thank you.

PR #2371, opened 2023-09-06
No empty description for polymorphic subtypes

### LINKED ISSUE #2370: Empty schema description if polymorphic subtype missing Javadoc
**Describe the bug**

- springdoc-openapi-javadoc causes an empty schema description to be generated for classes that are polymorphic subtypes and missing a Javadoc comment.

**To Reproduce**
Steps to reproduce the behavior: See linked pull request (#2371)

**Expected behavior**

- Schema description should remain null if class Javadoc is missing.

**Additional context**
Probably has the same cause as #2035 .

westse:
- Fixes #2370

### DIFF
--- a/springdoc-openapi-tests/springdoc-openapi-javadoc-tests/src/test/java/test/org/springdoc/api/app170/Cat.java
+++ b/springdoc-openapi-tests/springdoc-openapi-javadoc-tests/src/test/java/test/org/springdoc/api/app170/Cat.java
@@ -1,0 +1,22 @@
+package test.org.springdoc.api.app170;
+
+
+import io.swagger.v3.oas.annotations.media.Schema;
+
+@Schema
+public class Cat implements Animal {
+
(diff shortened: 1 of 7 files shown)


springdoc-openapi-tests/springdoc-openapi-javadoc-tests/src/test/java/test/org/springdoc/api/app170/Cat.java
```diff
@@ -0,0 +1,22 @@
+package test.org.springdoc.api.app170;
+
+
+import io.swagger.v3.oas.annotations.media.Schema;
+
+@Schema
+public class Cat implements Animal {
```
westse, line 7: Notice missing class Javadoc (and no explicit description set on `@Schema`), which should translate to null `Schema.description`

bnasslahsen:
@westse, 

PR #2385, opened 2023-09-21
support get javadoc description from getter method

eshizhan:
The API return [jpa projections](https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#projections.interfaces) is a typical application scenario.


eshizhan:
new PR #2387

PR #2394, opened 2023-09-29
Fix environment dependent unit test

zorglube:


bnasslahsen:
@zorglube,

It should be already fixed by #2366

Issue #2401, opened 2023-10-12
Enhanced Schema Configuration via Properties File and Template Generation


anton-tkachenko:
** Is your feature request related to a problem? Please describe. ** 

I have a need to configure most text fields within a schema through a properties file rather than using annotations. This will offer more flexibility and control, including:

Reliance on conventional property naming instead of specifying property values in the @Schema annotation.
Control over various text fields like endpoint summaries (using operation IDs), API descriptions, model names, titles, descriptions, and examples.
The current approach using springdoc.api-docs.resolve-schema-properties is limited, and I want a more comprehensive solution.

I can't modify the generated schema directly because it's not the source of truth, and any code changes would disrupt my customizations.

**  Describe the solution you'd like ** 

I propose the addition of a feature that generates a "template" properties file that corresponds to the project's schema. This template can then be easily populated with the desired content, removing the need to guess conventions when configuring the schema properties.

**  Describe alternatives you've considered ** 

I've explored potential alternatives, but I haven't found any built-in solutions that meet my requirements. To address this need, I plan to create an OpenApiCustomizer for this purpose. My ultimate goal is to contribute this solution to the springdoc project. However, I would greatly appreciate feedback and ideas from the community to ensure a better alignment with the project's goals.

Example Properties:

Let's consider a scenario where you want to configure the API documentation for a RESTful service. You might define properties like these in the template properties file:

properties
```
# API Description
springdoc.api-docs.description=This is a sample API for demonstration purposes.

# Model Information
springdoc.model.user.name=User
springdoc.model.user.title=User Model
springdoc.model.user.description=This model represents a user entity.
springdoc.model.user.example=...

# Endpoint Information
springdoc.endpoint.getUsers.summary=Get a list of users
springdoc.endpoint.getUsers.description=This endpoint retrieves a list of all users.
springdoc.endpoint.getUserById.summary=Get user by ID
springdoc.endpoint.getUserById.description=Retrieve a user by their unique identifier.
```

These properties allow you to control various aspects of the API documentation, including descriptions, model details, and endpoint summaries.

Sample YAML Schema:

Suppose you have the following sample YAML schema for the API:
```
yaml
Copy code
openapi: 3.0.0
info:
  title: Sample API
  version: 1.0.0
  description: This is a sample API for demonstration purposes.

paths:
  /users:
    get:
      summary: Get a list of users
      description: This endpoint retrieves a list of all users.
      responses:
        '200':
          description: Successful response
          content:
[...]

PR #2418, opened 2023-11-03
Add support for externalizing strings in generated openapi schema via…

tkachenkoas:
Originates from https://github.com/springdoc/springdoc-openapi/issues/2401

Can you please hint, what files should be adjusted to publish documentation properly?

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/configuration/SpringDocSpecificationStringPropertiesConfiguration.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/configuration/SpringDocSpecificationStringPropertiesConfiguration.java
@@ -25,6 +25,8 @@
 package org.springdoc.core.configuration;
 
 import org.springdoc.core.customizers.SpecificationStringPropertiesCustomizer;
+import org.springdoc.core.models.GroupedOpenApi;
+import org.springframework.beans.factory.config.BeanPostProcessor;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
@@ -32,6 +34,8 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Lazy;
 import org.springframework.core.env.PropertyResolver;
+
+import java.util.List;
 
 /**
  * The type Spring doc specification string properties configuration.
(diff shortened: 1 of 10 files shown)


tkachenkoas:
@bnasslahsen please take a look at this PR and let me know if something requires adjustment

bnasslahsen:
@tkachenkoas,

Thank you for your contribution. I believe it's not yet complete.

Do you think you can add the case of OpenAPI groups as well ?

tkachenkoas:
> @tkachenkoas,
> 
> Thank you for your contribution. I believe it's not yet complete.
> 
> Do you think you can add the case of OpenAPI groups as well ?

By "not complete" you mean all fields in the Schema model (like min/max etc.). I didn't plan to add them in the first version, but if you think it makes sense - surely, I'll add it

What case and behavior do you mean by supporting groups? 

E.g. sth like 
```springdoc.specification-strings.{group-name}.paths.{operationId}.description```
should override "general" string
```springdoc.specification-strings.paths.{operationId}.description```

I guess yes, it makes sense when we have identical method names "like getAccount" in UI api and MOBILE api (and in global schema one of them will become ```getAccount_1```)?..

**Comment by bnasslahsen** [comment]
@tkachenkoas,

Sorry for my late response. 
I confirm your properties description, is the exact expected result.

For the method names, more than 99% of cases they are unique per path and it can be considered for the first MVP.
Supporting polymorphism of exposed methods in controller, will bring more complexity and can be considered for another enhancment.

**Comment by tkachenkoas** [comment]
> @tkachenkoas,
> 
> Sorry for my late response. I confirm your properties description, is the exact expected result.
> 
> For the method names, more than 99% of cases they are unique per path and it can be considered for the first MVP. Supporting polymorphism of exposed methods in controller, will bring more complexity and can be considered for another enhancment.

ok, now it's my turn to respond with delay... :)

do I need to modify some other places for this functionality to appear in the documentation?

**Comment by bnasslahsen** [comment]
@tkachenkoas,

It doesn't seem to me that the groups support is implemented.
if it's the case, can you add the the corresponding another test case using ?
```properties
springdoc.specification-strings.{group-name}.paths.{operationId}.description
```

tkachenkoas:
> @tkachenkoas,
> 
> It doesn't seem to me that the groups support is implemented. if it's the case, can you add the the corresponding another test case using ?
> 
> ```ini
> springdoc.specification-strings.{group-name}.paths.{operationId}.description
> ```

ok, I guess I misunderstood one of the previous messages. Adjusted the logic and added a test-case

bnasslahsen:
@tkachenkoas,

Your push was too fast.
i will have to take care of it.

bnasslahsen:
@tkachenkoas,

code merged:
- Removed `springdoc.api-docs.specification-string-properties` property
- Replaced `springdoc.specification-strings` by `springdoc.spec-properties`

PR #2437, opened 2023-11-30
Bump swagger-core to version 2.2.19. Fixes #2360

### LINKED ISSUE #2360: Schema.RequiredMode.Required not working for spring mvc and kotlin
**Describe the bug**

When building the same case as the springdoc-kotlin-test module does [here](https://github.com/springdoc/springdoc-openapi/blob/main/springdoc-openapi-tests/springdoc-openapi-kotlin-tests/src/test/resources/results/app9.json) in a spring boot 3 webMVC app,  the openAPI-spec ends up with all the fields being reported as required.

**To Reproduce**
I've copied the resources from the test referenced above into an example app with a (failing) test which can be found here:
https://github.com/olbpetersson/springdoc-webmvc-test/blob/main/src/test/kotlin/com/example/demo/DemoApplicationTests.kt

and can be executed via `./gradlew bootRun & ./gradlew test`

**Expected behavior**
The fields annotated with @Schema.RequiredMode.NOT_REQUIRED should not end up in the required list. The openAPI-spec should be the same as the one in app9.json referenced above

**Screenshots**
-
**Additional context**
-

viktor-berglund-mw:
The problem mentioned in [this issue](https://github.com/springdoc/springdoc-openapi/issues/2360)  was caused by a bug in swagger core. It's fixed in version [2.2.19](https://github.com/swagger-api/swagger-core/releases/tag/v2.2.19) "give precedence to requiredMode annotation".

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,7 +61,7 @@
 		<maven-gpg-plugin.version>1.6</maven-gpg-plugin.version>
 		<maven-release-plugin.version>2.5.3</maven-release-plugin.version>
 		<nexus-staging-maven-plugin>1.6.8</nexus-staging-maven-plugin>
-		<swagger-api.version>2.2.17</swagger-api.version>
+		<swagger-api.version>2.2.19</swagger-api.version>
 		<swagger-ui.version>5.9.0</swagger-ui.version>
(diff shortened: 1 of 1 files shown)


bnasslahsen:
@viktor-berglund-mw,

Thank you for your contribution to the project

PR #2443, opened 2023-12-04
Respect schema annotations when using spring mvc with kotlin

viktor-berglund-mw:
This PR fixes a bug when using kotlin and spring mvc. The field `noRequiredNoNullableDefault` in `springdoc-openapi-tests/springdoc-openapi-kotlin-mvc-tests/src/test/kotlin/test/org/springdoc/api/app9/DemoController.kt` will be marked as required when it should not, it has a default value in the swagger annotation.

When adding `AnnotationIntrospectors`, the order matters. The fact that the `SpringDocJacksonKotlinModuleConfiguration` was dependent on `ObjectMapperProvider` and `SpringDocConfiguration` made it so that the `SpringDocRequiredModule` was added before the `KotlinModule`.

To fix this bug we create a primary bean for the `ObjectMapperProvider` when `KotlinModule` is on the class path to ensure that this module is added before `SpringDocRequiredModule`.

Tests are added for spring mvc using kotlin

Co-authored-by: Ola Petersson <olbpetersson@gmail.com>
Co-authored-by: Simon Lindhen <simon.lindhen@gmail.com>


viktor-berglund-mw:
@bnasslahsen do you think you will accept this PR for the next release? If not, is there something we can do to make that happen?

bnasslahsen:
@viktor-berglund-mw,

i will add you PR for the next release.

Issue #2477, opened 2024-01-08
buildApiResponses ignores produced ContentType in case of many @Operation


hexmind:
**Describe the bug**
one content type (a/a) is ignored

**To Reproduce**

- What is the actual and the expected result using OpenAPI Description (yml or json)?
input:
```
    @Operation(description = "a")
    @GetMapping(path = "/", produces = "a/a")
    public String a() {
        return "A";
    }

    @Operation(description = "b")
    @GetMapping(path = "/", produces = "b/b")
    public String b() {
        return "B";
    }
```
actual result:
```
            "content":{
              "b/b":{
                "schema":{
                  "type":"string"
                }
              }
            }
````

- Provide with a sample code (HelloController) or Test that reproduces the problem
https://github.com/springdoc/springdoc-openapi/pull/2478

A bug and solution are probably in `GenericResponseService.java`:
https://github.com/springdoc/springdoc-openapi/pull/2478/files#diff-bde5de3f4d4588ddcb3a1efb697aa8fc5fd857d63bc4e1d003d10d55d1e973cf

**Expected behavior**

```
        "responses":{
          "200":{
            "description":"OK",
            "content":{
              "b/b":{
                "schema":{
                  "type":"string"
                }
              }
              ,"a/a":{
                "schema":{
                  "type":"string"
                }
              }
            }
          }
        }

```

hexmind:
?

bnasslahsen:
@hexmind,

your PR is in draft mode. have completed your PR ?

hexmind:
Hi, my intention of this PR was to show a problem and a behavior only.

bnasslahsen:
@hexmind,

ok, will take it in consideration for further investigations.

PR #2478, opened 2024-01-08
#2477 buildApiResponses ignores produced ContentType in case of many @Operation

hexmind:


bnasslahsen:
closed ad no activity

PR #2511, opened 2024-02-15
Only exported association properties should be ignored in schema

### LINKED ISSUE #2295: Spring Data REST: force openapi schema export
When working with SDR, I need that the associated resources are expanded in the returned JSON. I have achieved this by adding @RestResource(exported = false) to the associated resource. For example:

```
@NotNull
@RestResource(exported = false)
@ManyToOne(fetch = FetchType.EAGER)
private Client client;

```
So far so good, as I can read and write to the association without extra requests. The problem is that I use openapi-generator to create feign clients of the API, and in this case the Client entity is not exported in the openapi schema, so I need this workaround to get it exported:

```
public Client getClientInline()
{
    return client;
}
```

I'd like to know if is there any way to get rid of this workaround. I made a test with an excerpt projection, but the schema also doesn't get exported. Adding @SchemaProperty also doesn't helps.

This is the current schema output:
```
"EntityModelIntegrationConfig": {
"required": [
"client",
"cron",
"enabled",
"template"
],
"type": "object",
"properties": {
"id": {
"type": "string",
"format": "uuid",
"readOnly": true
},
"cron": {
"type": "string"
},
"enabled": {
"type": "boolean"
},
"templateInline": {
"$ref": "#/components/schemas/Template"
},
"clientInline": {
"$ref": "#/components/schemas/Client"
},
"links": {
"type": "array",
"items": {
"$ref": "#/components/schemas/Link"
}
}
}
}
```
The expected behavior is to have client and template to be treated as properties, so I can remove the additional "inline" getters.

Sorry to post this here, but I didn't figured this out by looking at documentation / source code / stackoverflow

Thanks in advance

clementdenis:
Fixes #2295

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/utils/SpringDocDataRestUtils.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/utils/SpringDocDataRestUtils.java
@@ -55,6 +55,7 @@ import org.springframework.data.mapping.PersistentEntity;
 import org.springframework.data.mapping.PersistentProperty;
 import org.springframework.data.mapping.SimpleAssociationHandler;
 import org.springframework.data.mapping.context.PersistentEntities;
+import org.springframework.data.rest.core.annotation.RestResource;
 import org.springframework.data.rest.core.config.RepositoryRestConfiguration;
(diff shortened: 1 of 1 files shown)


bnasslahsen:
already fixed by https://github.com/springdoc/springdoc-openapi/pull/2501

PR #2579, opened 2024-04-19
fix: add support for leading tab characters with `trim-kotlin-indent`.

testersen:
We're using Springdoc with Kotlin + Ktlint. We have configured Ktlint to use tabs instead of spaces as indents.

```kt
	@GetMapping("/assets")
	@Operation(
		operationId = "getAssets",
		summary = "Get the assets of the signed in user",
		description = """
			Get the assets owned by the currently signed in user.
		""",
		// ^^^^ Tab characters here turned into a code block.
	)
	@[WithExternalEndUser ScopesAllowed("...")]
	fun getAssets(
		@RequestParam(required = false, defaultValue = "nb-NO") locale: String,
	): List<AssetCard> = TODO()
```

This pull request appends on the condition in the `countLeadingSpaces` method where the count loop filters on spaces. The change makes the condition break on a non space or tab character.

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/utils/PropertyResolverUtils.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/utils/PropertyResolverUtils.java
@@ -143,7 +143,7 @@ public class PropertyResolverUtils {
 	private int countLeadingSpaces(String line) {
         int count = 0;
         for (char ch : line.toCharArray()) {
-            if (ch != ' ') break;
+            if (ch != ' ' && ch != '\t') break;
             count++;
         }
         return count;


testersen:
@bnasslahsen Heyo :) Any ETA for when we can look into merging this? Would you like to see any changes in this suggestion?

bnasslahsen:
@testersen,

The PR will be merged today.
It should be part of the next release, by the end of June in the latest.

testersen:
Thanks @bnasslahsen! I also noticed that the current implementation does also not trim whitespace in `@Schema` annotations in fields such as descriptions - would this be something I should open a pull request for here, or should I do that another place?

bnasslahsen:
@testersen,

Please feel free to propose another PR and should integrate it.

PR #2582, opened 2024-04-22
pass HttpHeaders to ServerBaseUrlCustomizer

mshima:


mshima:
@bnasslahsen how should I handle backward compatibility?

mshima:
`ServletServerHttpRequest` could be used instead of `HttpServletRequest` to provide a webflux/webmvc common api using `ServerHttpRequest` interface.

PR #2626, opened 2024-06-15
Replace `Page` schema with `PagedModel` when `pageSerializationMode` is set to `VIA_DTO`

### LINKED ISSUE #2625: support `@EnableSpringDataWebSupport(pageSerializationMode = EnableSpringDataWebSupport.PageSerializationMode.VIA_DTO)`

starting with spring-boot-3.3.0 / spring-data-commons-3.3.0, you can choose to use `PagedModel<T>` (part of spring-data's public api) as (default) response representation of spring-data's `Page<T>` interface.

ref. https://docs.spring.io/spring-data/commons/reference/repositories/core-extensions.html#core.web.page.config

adding `@EnableSpringDataWebSupport(pageSerializationMode = EnableSpringDataWebSupport.PageSerializationMode.VIA_DTO)` does not change springdoc's behavior - it still uses the `Page` interface to infer the response representation (which kinda makes sense given it is still the controller method return type)..

it would be nice if springdoc would support `PageSerializationMode.VIA_DTO`, e.g. by making it aware of `SpringDataWebSettings.pageSerializationMode` or by some means of configuration.

* works (although not always desirable):
  * using `PagedModel` explicitly (i.e. changing controller return type from `Page<T>` to `PagedModel<T>`) and returning `new PagedModel<>(myPage)`
* does not work:
  * adding `@ApiResponse(content = @Content(schema = @Schema(implementation = PagedModel.class)))` 
    * this does not respect generics - i.e. it results in one `PagedModel<?>` schema (`content: any[]`) instead of specific representations tied to the contained type like it works for `Page<T>` (e.g. `Page<Person>` -> `PagePerson` schema (`content: Person[]`))
  * using `SpringDocUtils.getConfig().replaceWithClass(Page.class, PagedModel.class)` 
    * this does not respect generics as well - same problem as above

**Additional context**

n/a

EvaristeGalois11:
Hi, in the last update Spring Data introduced an automatic way of converting `Page` to `PagedModel`, more info [here](https://github.com/spring-projects/spring-data-commons/issues/3024) and [here](https://docs.spring.io/spring-data/commons/reference/repositories/core-extensions.html#core.web.page).

With this PR Springdoc should be able to recognize the configuration `@EnableSpringDataWebSupport(pageSerializationMode = EnableSpringDataWebSupport.PageSerializationMode.VIA_DTO)` and act accordingly.

I had to bump the version of the spring parent to 3.3.0 because I needed the new class `PagedModel`. If you prefer to not clutter this PR with this upgrade I will split it off to a separate PR.

Close #2625 

Bye!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 	<parent>
 		<groupId>org.springframework.boot</groupId>
 		<artifactId>spring-boot-starter-parent</artifactId>
-		<version>3.2.4</version>
+		<version>3.3.0</version>
 	</parent>
 
 	<licenses>
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/configuration/SpringDocPageConfiguration.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/configuration/SpringDocPageConfiguration.java
@@ -1,0 +1,96 @@
+/*
+ *
+ *  *
+ *  *  *
+ *  *  *  * Copyright 2019-2024 the original author or authors.
+ *  *  *  *
+ *  *  *  * Licensed under the Apache License, Version 2.0 (the "License");
+ *  *  *  * you may not use this file except in compliance with the License.
+ *  *  *  * You may obtain a copy of the License at
+ *  *  *  *
+ *  *  *  *      https://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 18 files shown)


bnasslahsen:
@EvaristeGalois11,

PR #2627, opened 2024-06-15
Ensure compatibility with previous version of spring data

EvaristeGalois11:
First of all,  thank you very much for accepting my previous contribution :D

This is a follow up to https://github.com/springdoc/springdoc-openapi/pull/2626

I saw that you did some changes to it, in particular moving the bean declaration from a dedicated configuration class to the already present `SpringDocPageableConfiguration`.

This sadly breaks compatibility with spring data < 3.3.0 because without the `@ConditionalOnClass(PagedModel.class)` as a safeguard the `@ConditionalOnBean` alone isn't safe to use and a `ClassNotFoundException` is thrown when the class isn't on the classpath.

I tweaked a bit the bean declaration to make sure that it doesn't break anymore if `SpringDataWebSettings` isn't present.

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/configuration/SpringDocPageableConfiguration.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/configuration/SpringDocPageableConfiguration.java
@@ -43,6 +43,7 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Lazy;
 import org.springframework.data.domain.Pageable;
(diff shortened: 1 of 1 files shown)


bnasslahsen:
@EvaristeGalois11,

Did you test it with previous versions of spring data  < 3.3.0  ?

EvaristeGalois11:
@bnasslahsen Yep

I pushed the project I'm using to do some tests, it's a bit rough but if you want to double check it should be fine

Here it is https://github.com/EvaristeGalois11/paged-model-swagger

bnasslahsen:
Thank you @EvaristeGalois11 - the fix looks good!
It's now merged.

Issue #2645, opened 2024-07-12
`trim-kotlin-indent` function also applies to Schema annotation


e-build:
**Is your feature request related to a problem? Please describe.**
I added the `trim-kotlin-indent` feature in the [last PR](https://github.com/springdoc/springdoc-openapi/pull/2535). Mainly, we added logic to remove Kotlin indents for strings in the Operation annotation properties. After checking the issues that arose, I found that I wanted to apply the `@Schema` annotation as well ([PR](https://github.com/springdoc/springdoc-openapi/pull/2579)), so I'm going to work on it.

**Describe the solution you'd like**
I will also add logic to remove indentation for property values ​​of `@Schema`

GeorgEchterling:
I've previously implemented this as a `GlobalOpenApiCustomizer`:

```kotlin
@Configuration
class OpenApiConfig {
    /**
     * Customizes the OpenAPI specification that was generated from Swagger annotations.
     * @see formatAllTexts
     * @see reorderAllFields
     */
    @Bean
    fun openApiCustomizer() = GlobalOpenApiCustomizer { openapi ->
        formatAllTexts(openapi)
        reorderAllFields(openapi)
    }

    /**
     * Trims Kotlin's indent on relevant specification fields, since trimIndent cannot be used directly in annotations.
     * This way, we can comfortably use multiline strings in annotations without
     * either formatting errors or having to start at column 1.
     */
    private fun formatAllTexts(openapi: OpenAPI) {
        openapi.components?.schemas?.values?.forEach(this::formatSchema)
        openapi.paths?.values?.forEach { path ->
            path.description = path.description?.trimIndent()
            path.readOperations()?.forEach(this::formatOperation)
        }
        openapi.components?.securitySchemes?.values?.forEach { securityScheme ->
            // @SecurityScheme.descriptions
            securityScheme.description = securityScheme.description?.trimIndent()
        }
        openapi.tags?.forEach { tag ->
            // @Tag.description
            tag.description = tag.description?.trimIndent()
        }
    }

    /**
     * Recursively formats the [schema]'s descriptions and examples.
     */
    private fun formatSchema(schema: Schema<*>) {
        // @Schema.description
        schema.description = schema.description?.trimIndent()

        // @Schema.example
        val example: Any? = schema.example
        if (example is String) schema.example = example.trimIndent()

        // Properties on classes
        schema.properties?.values?.forEach(this::formatSchema)

        // @ArraySchema.schema
[...]

PR #2653, opened 2024-07-23
Trim indent apply schema description

e-build:
implement #2645

While looking for where to implement the logic to remove the indent of the schema annotation, I wrote it in `AbstractOpenApiResource`. I thought it would be easy to track the implementation according to the `trim-kotlin-indent` setting and add features during future maintenance.

If you think there is a more appropriate class where this implementation should be located, please let me know. I will fix it.


testersen:
Legendary! I was going to start working on this soon:tm: - time hasn't been on my side lately!

Thanks!

testersen:
Hello my legend @bnasslahsen, do you think it would be possible to merge this soon?

Thanks!

bnasslahsen:
Hi @testersen,

It will be available by the end of month with the next release.

testersen:
@bnasslahsen That's awesome! Thank you!

PR #2664, opened 2024-07-31
Refactor Replace hardcoded schema prefix length

zinzoddari:
HI 🙋🏻‍♀️

The code was previously hardcoding the length of the string [ `#/components/schemas/` ] (21) directly in the code. It has now been modified to use the length of the constant value declared in the [ `Components` ] class of [ io.swagger.v3.oas.models ].

Thank you 🙇🏻‍♀️


zinzoddari:
The reason for this change was that when I first looked at the code, it took me a while to understand why the number 21 was used. By relying on the constant value, any future changes to the value can be handled without manually finding and updating the code. Additionally, it makes it easier to understand which length of value is being referenced. 💪🏻😃

PR #2668, opened 2024-08-04
Replace swagger urls in `org.springdoc.core.properties.AbstractSwaggerUiConfigProperties#urls` only if url is changed. Test to check if the issue is fixed. Fixes #2509.

### LINKED ISSUE #2509: ArrayIndexOutOfBoundsException in SwaggerUiConfigParameters
**Description:**

An "ArrayIndexOutOfBoundsException" error has been reported within the SwaggerUiConfigParameters class. This exception occurs during the execution of methods invoked by the AbstractSwaggerIndexTransformer class. The issue was raised by one of our customers, and although we've attempted to reproduce it, we have been unsuccessful thus far. However, monitoring of all customer logs indicates that this issue is not isolated and has affected multiple customers.

**Methode:**

`protected String addParameters(String html) throws JsonProcessingException {
		String layout = swaggerUiConfigParameters.getLayout() != null ? swaggerUiConfigParameters.getLayout() : "StandaloneLayout";
		StringBuilder stringBuilder = new StringBuilder("layout: \"" + layout + "\" ,\n");

		Map<String, Object> parametersObjectMap = swaggerUiConfigParameters.getConfigParameters().entrySet().stream()
				.filter(entry -> !SwaggerUiConfigParameters.OAUTH2_REDIRECT_URL_PROPERTY.equals(entry.getKey()))
				.filter(entry -> !SwaggerUiConfigParameters.URL_PROPERTY.equals(entry.getKey()))
				.filter(entry -> !SwaggerUiConfigParameters.URLS_PROPERTY.equals(entry.getKey()))
				.filter(entry -> SwaggerUiConfigParameters.VALIDATOR_URL_PROPERTY.equals(entry.getKey())
						|| ((entry.getValue() instanceof String) ? StringUtils.isNotEmpty((String) entry.getValue()) : entry.getValue() != null))
				.collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (e1, e2) -> e2,
						LinkedHashMap::new));

		if (!CollectionUtils.isEmpty(parametersObjectMap)) {
			String result = objectMapper.writerWithDefaultPrettyPrinter().writeValueAsString(parametersObjectMap);
			result = result.substring(1, result.length() - 1);
			stringBuilder.append(result);
			html = html.replace("layout: \"StandaloneLayout\"", stringBuilder.toString());
		}

		return html;
	}`

**Stacktrace:**

`java.lang.ArrayIndexOutOfBoundsException: Index 11 out of bounds for length 11
at java.util.stream.SortedOps$SizedRefSortingSink.accept(Unknown Source)
at java.util.HashMap$KeySpliterator.forEachRemaining(Unknown Source)
at java.util.stream.AbstractPipeline.copyInto(Unknown Source)
at java.util.stream.AbstractPipeline.wrapAndCopyInto(Unknown Source)
at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(Unknown Source)
at java.util.stream.AbstractPipeline.evaluate(Unknown Source)
at java.util.stream.ReferencePipeline.collect(Unknown Source)
at org.springdoc.core.properties.SwaggerUiConfigParameters.put(SwaggerUiConfigParameters.java:307)
at org.springdoc.core.properties.SwaggerUiConfigParameters.getConfigParameters(SwaggerUiConfigParameters.java:284)
at org.springdoc.ui.AbstractSwaggerIndexTransformer.addParameters(AbstractSwaggerIndexTransformer.java:198)
at org.springdoc.ui.AbstractSwaggerIndexTransformer.defaultTransformations(AbstractSwaggerIndexTransformer.java:173)
`
**Possible Explanation:**
The ArrayIndexOutOfBoundsException typically occurs when attempting to access an element of an array or collection at an index that is beyond its bounds. In this context, the usage of the sorted operation in a stateful stream operation within the SwaggerUiConfigParameters class might lead to unexpected behavior. Stateful stream operations maintain state during the execution of the stream pipeline, which could potentially alter the order or length of elements being processed. Consequently, if the index being accessed within the SwaggerUiConfigParameters class exceeds the updated length of the array or collection due to the stateful operation, it could result in the ArrayIndexOutOfBoundsException being thrown.

Similar non springdoc-issue:
https://github.com/oracle/graal/issues/5902

**To Reproduce**
Unable to reproduce locally

org.springdoc:springdoc-openapi-starter-webmvc-ui:2.2.0
org.springdoc:springdoc-openapi-starter-webmvc-ui:2.2.0
Index page doesn't load
Unable to reproduce
**Expected behavior**

- A clear and concise description of what you expected to happen.
- What is the expected result using OpenAPI Description (yml or json)?

**Screenshots**

**Additional context**

dgswan:
Replace swagger urls in `org.springdoc.core.properties.AbstractSwaggerUiConfigProperties#urls` only if url is changed. Test to check if the issue is fixed. Fixes #2509.

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/ui/AbstractSwaggerWelcome.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/ui/AbstractSwaggerWelcome.java
@@ -33,6 +33,8 @@ import org.springdoc.core.properties.SwaggerUiConfigProperties;
 import org.springframework.util.CollectionUtils;
 import org.springframework.web.util.UriComponentsBuilder;
 
+import java.util.Objects;
+
 import static org.springdoc.core.utils.Constants.SWAGGER_UI_OAUTH_REDIRECT_URL;
 import static org.springdoc.core.utils.Constants.SWAGGER_UI_URL;
 import static org.springframework.util.AntPathMatcher.DEFAULT_PATH_SEPARATOR;
@@ -134,12 +136,17 @@ public abstract class AbstractSwaggerWelcome {
 			else
 				swaggerUiConfigParameters.addUrl(apiDocsUrl);
 			if (!CollectionUtils.isEmpty(swaggerUiConfig.getUrls())) {
-				swaggerUiConfig.cloneUrls().forEach(swaggerUrl -> {
-					swaggerUiConfigParameters.getUrls().remove(swaggerUrl);
-					if (!swaggerUiConfigParameters.isValidUrl(swaggerUrl.getUrl()))
-						swaggerUrl.setUrl(buildUrlWithContextPath(swaggerUrl.getUrl()));
-					swaggerUiConfigParameters.getUrls().add(swaggerUrl);
-				});
+				swaggerUiConfig.cloneUrls()
+						.stream()
+						.filter(swaggerUrl -> !swaggerUiConfigParameters.isValidUrl(swaggerUrl.getUrl()))
+						.forEach(swaggerUrl -> {
+							final var url = buildUrlWithContextPath(swaggerUrl.getUrl());
+							if (!Objects.equals(url, swaggerUrl.getUrl())) {
+								swaggerUiConfigParameters.getUrls().remove(swaggerUrl);
+								swaggerUrl.setUrl(url);
+								swaggerUiConfigParameters.getUrls().add(swaggerUrl);
+							}
+						});
 			}
 		}
 		calculateOauth2RedirectUrl(uriComponentsBuilder);
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/ui/AbstractSwaggerWelcome.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/ui/AbstractSwaggerWelcome.java
@@ -142,9 +142,10 @@ public abstract class AbstractSwaggerWelcome {
 						.forEach(swaggerUrl -> {
 							final var url = buildUrlWithContextPath(swaggerUrl.getUrl());
 							if (!Objects.equals(url, swaggerUrl.getUrl())) {
-								swaggerUiConfigParameters.getUrls().remove(swaggerUrl);
-								swaggerUrl.setUrl(url);
-								swaggerUiConfigParameters.getUrls().add(swaggerUrl);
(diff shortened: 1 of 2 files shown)


springdoc-openapi-starter-common/src/main/java/org/springdoc/ui/AbstractSwaggerWelcome.java
```diff
@@ -134,12 +136,17 @@ else if (swaggerUiConfigParameters.isValidUrl(swaggerUiUrl))
...
-						swaggerUrl.setUrl(buildUrlWithContextPath(swaggerUrl.getUrl()));
-					swaggerUiConfigParameters.getUrls().add(swaggerUrl);
-				});
+				swaggerUiConfig.cloneUrls()
+						.stream()
+						.filter(swaggerUrl -> !swaggerUiConfigParameters.isValidUrl(swaggerUrl.getUrl()))
+						.forEach(swaggerUrl -> {
+							final var url = buildUrlWithContextPath(swaggerUrl.getUrl());
+							if (!Objects.equals(url, swaggerUrl.getUrl())) {
+								swaggerUiConfigParameters.getUrls().remove(swaggerUrl);
```
dssievewright, line 145: In some scenarios, I believe this will still encounter the reported ArrayIndexOutOfBoundsException in https://github.com/springdoc/springdoc-openapi/issues/2509. I can duplicate that in the test file by adding `@TestPropertySource(properties = {"spring.mvc.servlet.path=test"})` to the class and you'll hit both ArrayIndexOutOfBoundsExceptions and ConcurrentModificationExceptions depending on the thread.

Would it make more sense to do 
`swaggerUiConfigParameters.getUrls().stream().filter(o -> o.equals(swaggerUrl)).forEach(subUrl -> subUrl.setUrl(url));`
instead?

dgswan: Good point, you are right, I missed it. It would be safer to avoid mutation of the collection at all, instead only the url itself should be mutated when necessary. Thank you! Fixed in https://github.com/springdoc/springdoc-openapi/pull/2668/commits/570e367b4ad4b51cb2efc80815b3a97aad8cba74.

dssievewright (approved):

PR #2760, opened 2024-10-11
Added support for RequestBody as a meta-annotation

### LINKED ISSUE #2759: Support RequestBody as a meta-annotation
**Problem:**
 
`io.swagger.v3.oas.annotations.parameters.RequestBody` has three targets it can be used on:
- METHOD
- PARAMETER
- ANNOTATION_TYPE
 
When springdoc resolves annotation in `AbstractRequestService.isRequestBodyParam()`, it checks if there's such annotation on a parameter directly:
 
```
methodParameter.getParameterAnnotation(RequestBody.class) != null
```
 
or on a method:
```
AnnotatedElementUtils.findMergedAnnotation(methodParameter.getMethod(), RequestBody.class) != null
```
 
but it does not cover `ANNOTATION_TYPE`, so any annotation that would use `RequestBody` as a meta-annotation won't work. This is my case, as I have a special annotation for the request body that is not native to Spring.
 
**Workaround:**
 
Use two annotations:
```
@MyRequestBody @io.swagger.v3.oas.annotations.parameters.RequestBody
```
I use full path because the name is the same between one in  Swagger and one in Spring, and there are mix-ups.
 
**Solution:**
 
Check if `RequestBody` is a meta-annotation:
```
AnnotatedElementUtils.findMergedAnnotation(methodParameter.getParameter(), RequestBody.class) != null
```
This should cover both `PARAMETER` and `ANNOTATION_TYPE` cases.

edudar:
The change uses advanced `AnnotatedElementUtils` to find annotation instead of simple `getParameterAnnotation`.

Resolves #2759

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/service/AbstractRequestService.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/service/AbstractRequestService.java
@@ -783,8 +783,8 @@ public abstract class AbstractRequestService {
 	 * @return the boolean
  	 */
 	private boolean checkRequestBodyAnnotation(MethodParameter methodParameter) {
-		return methodParameter.getParameterAnnotation(org.springframework.web.bind.annotation.RequestBody.class) != null
-				|| methodParameter.getParameterAnnotation(io.swagger.v3.oas.annotations.parameters.RequestBody.class) != null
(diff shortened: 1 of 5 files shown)


bnasslahsen:
@edudar,

Can you check the failing build ?

edudar:
Fixed build and added tests

Issue #2691, opened 2024-08-29
Not distinction between different parameter conditions on the same path in swagger json.


Jelicho:
**Describe the bug**

Not distinction between different parameter conditions on the same path in swagger json.

**To Reproduce**

Simple example project
https://github.com/Jelicho/springdoc-openapi-parameter-condition-example

As in the example below, two separate APIs are provided with different parameter conditions on the same path.
The response format is also different. However, in swagger json and ui, they are configured and exposed as one.

- path : `/api/example/{exampleNo}`
- parameter condition : `application=service1`, `application=service2`
- Response : `ExampleDataForService1Dto`, `ExampleDataForService2Dto`

```java
@Operation(
	operationId = "exampleForService1",
	description = "Returns Example data for service1 application."
)
@GetMapping(value = "/{exampleNo}", params = "application=service1")
public ExampleDataForService1Dto exampleForService1(
	@PathVariable String exampleNo
) {
	return ExampleDataForService1Dto.builder()
		.exampleNo(exampleNo)
		.exampleName("exampleName")
		.build();
}

@Operation(
	operationId = "exampleForService2",
	description = "Returns Example data for service2 application."
)
@GetMapping(value = "/{exampleNo}", params = "application=service2")
public ExampleDataForService2Dto exampleForService2(
	@PathVariable String exampleNo
) {
	return ExampleDataForService2Dto.builder()
		.exampleNo(exampleNo)
		.exampleAddress("exampleAddress")
		.build();
}
```

<img width="1216" alt="image" src="https://github.com/user-attachments/assets/d4bcfac0-3ab0-4c07-ad2a-83bd590da2ea">
<details>
  <summary>swagger json</summary>

```json
// http://localhost:8080/v3/api-docs

{
  "openapi": "3.0.1",
  "info": {
    "title": "OpenAPI definition",
    "version": "v0"
  },
  "servers": [
    {
      "url": "http://localhost:8080",
      "description": "Generated server url"
    }
  ],
  "paths": {
    "/api/example/{exampleNo}": {
      "get": {
        "tags": [
          "Example"
        ],
        "description": "Returns Example data for service1 application.",
        "operationId": "exampleForService1_1",
        "parameters": [
          
        ],
        "responses": {
          "200": {
            "description": "OK",
            "content": {
              "*/*": {
                "schema": {
                  "$ref": "#/components/schemas/ExampleDataForService2Dto"
                }
              }
            }
          }
        }
      }
    }
  },
[...]

PR #2761, opened 2024-10-12
Add paths with params condition option

Jelicho:
@bnasslahsen 

Hello, I propose a new option to expose params condition in the path, allowing the value to be differently exposed in the path as well.

I understand the comment you provided regarding this issue (#2691 )and respect the direction of using the path and HttpMethod as identifiers for API uniqueness.

While it's true that params condition alone may not guarantee uniqueness, it plays a role in refining the request, making it a valid option to offer. Additionally, the Spring Framework's RequestMappingHandler supports API mapping separation based on params condition, and I believe there are cases where others have similar requirements.

Therefore, I would like to gradually improve this feature by providing it as an option that users can use if needed.

Of course, this is just another direction I'm suggesting.

Thank you.

resolved : #2691


bnasslahsen:
@Jelicho,

I would be happy to merge it, but you need to add the unit tests and the webflux implementation.

bnasslahsen:
@Jelicho,

This PR, will be closed as missing the unit tests.

PR #2777, opened 2024-11-08
Add SortAsQueryParam annotation

ondrejkrpec:
Hello,

I would like to propose a new annotation: `@SortAsQueryParam`. This annotation allows the use of the `Sort` object in query parameters. Currently, the project supports using the `Pageable` object in query parameters through the `@PageableAsQueryParam` annotation.

However, there isn't an easy way to utilize the Sort object independently without involving pagination. This limitation means that developers who need sorting capabilities without pagination have to implement workarounds.

By introducing the `@SortAsQueryParam` annotation, we can provide users with a straightforward option to include sorting in their API endpoints without the need for pagination. This enhancement will offer greater flexibility and convenience for those who require sorting functionality on its own.

Thank you for considering this proposal.


bnasslahsen:
@ondrejkrpec,

Thanks for your contribution to the project!

PR #2788, opened 2024-11-24
Obey annotations when flattening ParameterObject fields. Fixes #2787

### LINKED ISSUE #2787: Nested fields in ParameterObject do not take into account annotations on parent classes
## Describe the bug

For a `@ParameterObject` annotated value, any annotations on the fields of the target class are not taken into account when creating the flattened values for the parameters. I believe this leads to 3 types or incorrect definitions being produced:

### Hidden parent fields
When a field in the `@ParameterObject` class is defined as hidden, or any subfields are hidden, those fields and any child fields should not included in the generated schema.

For the following endpoint definition and model
```java
    @GetMapping("/hidden-parent")
    public void nestedParameterObjectWithHiddenParentField(@ParameterObject ParameterObjectWithHiddenField parameters) {

    }

    public record ParameterObjectWithHiddenField(
        @Schema(hidden = true) NestedParameterObject schemaHiddenNestedParameterObject,
        @Parameter(hidden = true) NestedParameterObject parameterHiddenNestedParameterObject,
        NestedParameterObject visibleNestedParameterObject
    ) {

    }

    public record NestedParameterObject(
        String parameterField) {
    }
```

I'd expect the `schemaHiddenNestedParameterObject` and `parameterHiddenNestedParameterObject` fields, plus any child fields, to be excluded from the generated result, e.g.

```json
      "parameters": [
          {
            "name": "visibleNestedParameterObject.parameterField",
            "in": "query",
            "required": false,
            "schema": {
              "type": "string"
            }
          }
        ]
```
but they're currently included, e.g.
```json
      "parameters": [
          {
            "name": "schemaHiddenNestedParameterObject.parameterField",
            "in": "query",
            "required": false,
            "schema": {
              "type": "string"
            }
          },
          {
            "name": "parameterHiddenNestedParameterObject.parameterField",
            "in": "query",
            "required": false,
            "schema": {
              "type": "string"
            }
          },
          {
            "name": "visibleNestedParameterObject.parameterField",
            "in": "query",
            "required": false,
            "schema": {
              "type": "string"
            }
          }
        ]
```

### Renamed parent fields
Where a field in the `@ParameterObject` is specifically given a name through an appropriate annotation, this name should be applied in the flattened definition, but the field name as defined in the code is currently used.

For the following endpoint definition and model
```java
    @GetMapping("/renamed-parent")
    public void nestedParameterObjectWithRenamedParentField(@ParameterObject ParameterObjectWithRenamedField parameters) {

    }

    public record ParameterObjectWithRenamedField(
        @Schema(name = "schemaRenamed") NestedParameterObject schemaRenamedNestedParameterObject,
        @Parameter(name = "parameterRenamed") NestedParameterObject parameterRenamedNestedParameterObject,
        NestedParameterObject originalNameNestedParameterObject
    ) {

    }

    public record NestedParameterObject(
        String parameterField) {
    }
```
I'd expect the `schemaRenamedNestedParameterObject` and `parameterRenamedNestedParameterObject` to be given the names from the annotation in the resulting definition, e.g.
```json
        "parameters": [
          {
            "name": "schemaRenamed.parameterField",
            "in": "query",
            "required": false,
            "schema": {
              "type": "string"
            }
          },
          {
            "name": "parameterRenamed.parameterField",
            "in": "query",
            "required": false,
            "schema": {
              "type": "string"
            }
          },
          {
            "name": "originalNameNestedParameterObject.parameterField",
            "in": "query",
            "required": false,
            "schema": {
              "type": "string"
            }
          }
        ]
```
but the original field names are used, e.g.
```json
        "parameters": [
          {
            "name": "schemaRenamedNestedParameterObject.parameterField",
            "in": "query",
            "required": false,
            "schema": {
              "type": "string"
            }
          },
          {
            "name": "parameterRenamedNestedParameterObject.parameterField",
            "in": "query",
            "required": false,
            "schema": {
              "type": "string"
            }
          },
          {
            "name": "originalNameNestedParameterObject.parameterField",
            "in": "query",
            "required": false,
            "schema": {
              "type": "string"
            }
          }
        ]
```

###  Mandatory parameters in non-mandatory parent fields
Where a fields is marked as required (or resolves as required for a RequiredMode.AUTO annotation), but the parent fields is not required, that child field would not be mandatory within the request. There is a challenge here that the field would be required if any of the other fields from the parent are set, but there's no way of defining that in a flattened structure, so the definition should default to not forcing a field that's only conditionally required.

Given the following endpoint and model
```java
    @GetMapping("/optional-parent")
    public void nestedParameterObjectWithOptionalParentField(@ParameterObject ParameterObjectWithOptionalField parameters) {

    }

    public record ParameterObjectWithOptionalField(
        @Schema(requiredMode = Schema.RequiredMode.NOT_REQUIRED) NestedRequiredParameterObject schemaNotRequiredNestedParameterObject,
        @Parameter NestedRequiredParameterObject parameterNotRequiredNestedParameterObject,
        @Parameter(required = true) NestedRequiredParameterObject requiredNestedParameterObject
    ) {

    }

    public record NestedRequiredParameterObject(
        @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String requiredParameterField) {
    }
```
I'd expect the fields makes as REQUIRED on an object that isn't set as REQUIRED to be shown as not required in the flattened definition, e.g.
```json
        "parameters": [
          {
            "name": "schemaNotRequiredNestedParameterObject.requiredParameterField",
            "in": "query",
            "required": false,
            "schema": {
              "type": "string"
            }
          },
          {
            "name": "parameterNotRequiredNestedParameterObject.requiredParameterField",
            "in": "query",
            "required": false,
            "schema": {
              "type": "string"
            }
          },
          {
            "name": "requiredNestedParameterObject.requiredParameterField",
            "in": "query",
            "required": true,
            "schema": {
              "type": "string"
            }
          }
        ]
```
but the definition currently only takes into account the child element annotations to treats the fields as required, e.g.
```json
        "parameters": [
          {
            "name": "schemaNotRequiredNestedParameterObject.requiredParameterField",
            "in": "query",
            "required": true,
            "schema": {
              "type": "string"
            }
          },
          {
            "name": "parameterNotRequiredNestedParameterObject.requiredParameterField",
            "in": "query",
            "required": true,
            "schema": {
              "type": "string"
            }
          },
          {
            "name": "requiredNestedParameterObject.requiredParameterField",
            "in": "query",
            "required": true,
            "schema": {
              "type": "string"
            }
          }
        ]
```

## Additional Information
* The version of Spring Boot being used does not seem to have any impact. The above examples were generated using 3.4.0 and the latest SNAPSHOT of springdoc-openapi from e8de90ee6b33cfb9a1c8539294a645b82d00a51d on the `main` branch.

mc1arke:
When creating the flattened parameter definitions for an item annotated with `@ParameterObject`, the `@Schema` and `@Property` annotations on any fields prior to the final child-node of each branch of the parameter tree are not taken into consideration. This results in the field name in the code being used even where annotations may override that, prevents the fields being hidden where one of the parent fields is marked as hidden but the child field isn't hidden, and marks a field as mandatory even where the field would only be mandatory if the parent object had been declared through a sibling of the target field being set. To overcome this, whilst the flattened parameter map is being built, each field is now being inspected for a `@Parameter` annotation or - where the `@Parameter` annotation isn't found - a `@Schema` annotation. Where a custom name is included in either of those annotations then the overridden field name is used in the flattened definition. If either annotation declares the field as hidden then the field and all child fields are removed from the resulting definition, and a field is no longer set as mandatory unless the parent field was also declared as mandatory, or resolved as non-null and was set in an automatic state.

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/extractor/DelegatingMethodParameter.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/extractor/DelegatingMethodParameter.java
@@ -94,6 +94,11 @@ public class DelegatingMethodParameter extends MethodParameter {
 	private final Annotation[] methodAnnotations;
 
 	/**
+	 * The annotations to mask from the list of annotations on this method parameter.
+	 */
+	private final Annotation[] maskedAnnotations;
+
+	/**
 	 * The Is not required.
 	 */
 	private boolean isNotRequired;
@@ -105,17 +110,19 @@ public class DelegatingMethodParameter extends MethodParameter {
 	 * @param parameterName                  the parameter name
 	 * @param additionalParameterAnnotations the additional parameter annotations
 	 * @param methodAnnotations              the method annotations
+	 * @param maskedAnnotations              any annotations that should not be included in the final list of annotations
 	 * @param isParameterObject              the is parameter object
 	 * @param isNotRequired                  the is required
 	 */
-	DelegatingMethodParameter(MethodParameter delegate, String parameterName, Annotation[] additionalParameterAnnotations, Annotation[] methodAnnotations, boolean isParameterObject, boolean isNotRequired) {
+	DelegatingMethodParameter(MethodParameter delegate, String parameterName, Annotation[] additionalParameterAnnotations, Annotation[] methodAnnotations, Annotation[] maskedAnnotations, boolean isParameterObject, boolean isNotRequired) {
 		super(delegate);
 		this.delegate = delegate;
 		this.additionalParameterAnnotations = additionalParameterAnnotations;
 		this.parameterName = parameterName;
 		this.isParameterObject = isParameterObject;
 		this.isNotRequired = isNotRequired;
-		this.methodAnnotations =methodAnnotations;
+		this.methodAnnotations = methodAnnotations;
+		this.maskedAnnotations = maskedAnnotations;
 	}
 
 	/**
@@ -146,7 +153,7 @@ public class DelegatingMethodParameter extends MethodParameter {
 			}
 			else {
 				String name = pNames != null ? pNames[i] : p.getParameterName();
-				explodedParameters.add(new DelegatingMethodParameter(p, name, null, null, false, false));
+				explodedParameters.add(new DelegatingMethodParameter(p, name, null, null, null, false, false));
 			}
 		}
 		return explodedParameters.toArray(new MethodParameter[0]);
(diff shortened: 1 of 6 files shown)


bnasslahsen:
@mc1arke,

PR #2832, opened 2024-12-31
Obey annotations when flattening ParameterObject fields. Fixes #2817

### LINKED ISSUE #2817: Automatically add required if a field is @notNull or @NotBlank
I have a project with Swagger 3, I noticed a different behavior compared to Swagger 2 and I think it's a bug because with a structure of a certain type, I can make the call to the API, but Swagger doesn't allow me to do it.

I have this API:
```
    @PostMapping(path = "/testOne", consumes = MediaType.APPLICATION_JSON_VALUE)
    ResponseEntity<String> testOne(@ParameterObject @Valid ClassOne test) {
        return ResponseEntity.ok("ok");
    }
```

```
@Data
public class ClassOne {

    @Valid
    private ClassTwo description;

}
```

```
@Data
@JsonIgnoreProperties(ignoreUnknown = true)
public class ClassTwo {

    @NotNull
    private String name;
}
```

As you can see, `ClassOne` has an object with the `@Valid` annotation and I am not obliged to pass it (because maybe I want to pass other attributes that I have not added in this example). However, the `@Valid` annotation ensures that if I pass any attribute that is inside that class, then it must respect the rules that I indicate inside (@NotNull etc.). The problem is that on the swagger it says required, even if I do not indicate required and therefore I cannot make the call to the API. This is therefore a problem as I said before, because I may want to send other attributes of `ClassOne`.
This problem occurs only in the case of `@PraameterObject`, while if I use `@RequestBody`, it is not mandatory.

In my opinion swagger should not automatically add requird if an attribute has the annotation `@NotNull` or `@NotBlank`, it should refer to what the user indicates with the annotation @Schema

attached is a zip with an example.
[demo.zip](https://github.com/user-attachments/files/18166665/demo.zip)

mc1arke:
When creating the flattened parameter definitions for an item annotated with @ParameterObject, the @Schema and @Property annotations on any fields prior to the final child-node of each branch of the parameter tree are not taken into consideration. This results in the field name in the code being used even where annotations may override that, prevents the fields being hidden where one of the parent fields is marked as hidden but the child field isn't hidden, and marks a field as mandatory even where the field would only be mandatory if the parent object had been declared through a sibling of the target field being set. To overcome this, whilst the flattened parameter map is being built, each field is now being inspected for a @Parameter annotation or - where the @Parameter annotation isn't found - a @Schema annotation. Where a custom name is included in either of those annotations then the overridden field name is used in the flattened definition. If either annotation declares the field as hidden then the field and all child fields are removed from the resulting definition, and a field is no longer set as mandatory unless the parent field was also declared as mandatory, or resolved as non-null and was set in an automatic state, or the child field is specifically set as required in the Schema or Parameter annotation.

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/extractor/DelegatingMethodParameter.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/extractor/DelegatingMethodParameter.java
@@ -94,6 +94,11 @@ public class DelegatingMethodParameter extends MethodParameter {
 	private final Annotation[] methodAnnotations;
 
 	/**
+	 * The annotations to mask from the list of annotations on this method parameter.
+	 */
+	private final Annotation[] maskedAnnotations;
+
+	/**
 	 * The Is not required.
 	 */
 	private boolean isNotRequired;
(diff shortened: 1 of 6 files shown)


bnasslahsen:
@mc1arke,

I have done the merge, with few simplifications, to make the code much readable.
All tests are passing. Do not hesitate to review it or add other test cases.

PR #2881, opened 2025-01-31
fix defaultValue when using @PageableDefault together with one-indexed-parameters

pheyken:
Hi,

we noticed that the default value for the page was 0 when having a spring-boot application with `spring.data.web.pageable.one-indexed-parameters=true` as well as using a `@PageableDefault` on the controller.

The `@PageableDefault` annotation has 0 as the default value for the page (see [here](https://github.com/spring-projects/spring-data-commons/blob/f86ee7f97f800b34487322a36baa165fc3a4a739/src/main/java/org/springframework/data/web/PageableDefault.java#L62)). This is the desired behavior (as far as I can tell) as spring normalizes the pageable before injecting it into the controller method. This is done in [this](https://github.com/spring-projects/spring-data-commons/blob/f86ee7f97f800b34487322a36baa165fc3a4a739/src/main/java/org/springframework/data/web/PageableHandlerMethodArgumentResolverSupport.java#L284) method.

So, e.g., if the newly added test controller was to be called with the following query parameters
```
/test1?prefix_pages=1
```

the injected pageable in the controller would return 0 for getPageNumber()
```
@GetMapping("/test1")
public String getPatientList1(
    @PageableDefault(size = 100, sort = { "someField", "someoTHER" }, direction = Sort.Direction.DESC) 
    @ParameterObject 
    Pageable pageable
) {
        System.out.println(pageable.getPageNumber()) // prints 0
	return "bla";
}
```

Let me know if you prefer to have a separate TestApp for this, happy to adjust this.

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/customizers/DataRestDelegatingMethodParameterCustomizer.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/customizers/DataRestDelegatingMethodParameterCustomizer.java
@@ -64,6 +64,9 @@ import org.springframework.data.web.SortDefault;
 
 /**
  * The type Data rest delegating method parameter customizer.
+ *
+ *  @author bnasslahsen
+ *  @author pheyken
  */
 public class DataRestDelegatingMethodParameterCustomizer implements DelegatingMethodParameterCustomizer {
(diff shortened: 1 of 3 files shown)


bnasslahsen:
@pheyken,

PR #2891, opened 2025-02-06
Refactor trimIndent Method

Ryan-Dia:
These changes aim to make the code more maintainable and easier to understand. Please review and let me know if any further improvements are needed.

📌 Code Refactoring 
   - Updated the method implementation to use Java Streams for cleaner and more concise code.
   - Enhanced exception handling with better logging to ensure stability if unexpected errors occur.
   - **String Concatenation Optimization**  
     - Replace the current use of reduce() with Collectors.joining() to optimize performance when joining strings.
       - why? Using reduce((a, b) -> a + newLine + b) for string concatenation can be inefficient because it creates a new String object every time. Instead, using Collectors.joining(newLine) is a more efficient approach as it internally utilizes a StringBuilder to handle the concatenation process. 
   - **Newline Handling Improvement**  
     - Use the regular expression "\\r?\\n" for splitting the input text to support both Unix and Windows newline formats.
     - Even if the output uses "\n" consistently, handling the input correctly is necessary for robustness across different environments.
   - The original code, which used a comprehensive try-catch block for the entire method, has been modified to apply exception handling only to parts where exceptions are likely to occur. This reduces unnecessary exception handling and improves both code readability and performance.

📌 javadoc Enhancement 
   - Revised the documentation to clearly state the behavior of the method.
   - Added details about how the method handles null input and specifies the return values.
   - Improved overall clarity to better communicate the intent and functionality of the method.

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/utils/PropertyResolverUtils.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/utils/PropertyResolverUtils.java
@@ -115,9 +115,11 @@ public class PropertyResolverUtils {
 	 * Returns a string where all leading indentation has been removed from each line.
 	 * It detects the smallest common indentation of all the lines in the input string,
 	 * and removes it.
+	 * If the input text is {@code null}, the method returns {@code null}.
 	 *
-	 * @param text The original string with possible leading indentation.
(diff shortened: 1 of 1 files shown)


bnasslahsen:
@Ryan-Dia,

PR #2896, opened 2025-02-08
Only filter out actuator endpoints with double asterisks. Fixes #2895

### LINKED ISSUE #2817: Automatically add required if a field is @notNull or @NotBlank

### LINKED ISSUE #2895: Endpoints with double-asterisk in the path are excluded from the Swagger output
**Describe the bug**

Since 2.7.0, a path that contains a double asterisk do not get included in the Swagger output. 

**To Reproduce**
Steps to reproduce the behaviour:

1. Define an endpoint with a double asterisk in the path matching
```java
@RestController
public class WildcardController {
    @PostMapping("/**")
    @Operation(summary = "My Wildcard Operation")
    public String getItem(HttpServletRequest request) {
        return request.getPathInfo();
    }
}
```
2. Check the swagger output

**Expected behaviour**

The endpoint to be included in the swagger output, e.g. 
```json
{
  "openapi": "3.1.0",
  "info": {
    "title": "OpenAPI definition",
    "version": "v0"
  },
  "servers": [
    {
      "url": "http://localhost",
      "description": "Generated server url"
    }
  ],
  "paths": {
    "/**": {
      "post": {
        "tags": [
          "wildcard-controller"
        ],
        "summary": "My Wildcard Operation",
        "operationId": "getItem",
        "responses": {
          "200": {
            "description": "OK",
            "content": {
              "*/*": {
                "schema": {
                  "type": "string"
                }
              }
            }
          }
        }
      }
    }
  },
  "components": {}
}
```

**Additional context**
https://github.com/springdoc/springdoc-openapi/commit/1a9a1d44f3be364dbde925fa2c6d5725c7dd1f4d appears to be the cause of this, which presumably was only supposed to filter out actuator endpoints that contained double asterisks, but is also filtering out non-actuator endpoints

mc1arke:
All endpoints that contain a double-asterisks in the path are currently being excluded from the output Swagger definition, rather than just excluding match-all actuator endpoints. The endpoint matching is therefore being altered to pre-filter the actuator endpoints to remove any that contain a double-asterisks in the path, and remove the subsequent filtering of double asterisks in the combined endpoints list so that non-actuator endpoints aren't dropped by the actuator filtering.

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java
@@ -661,8 +661,7 @@ public abstract class AbstractOpenApiResource extends SpecFilter {
 			}
 			
 			PathItem pathItemObject = buildPathItem(requestMethod, operation, operationPath, paths);
-			if (!StringUtils.contains(operationPath, "**"))
-				paths.addPathItem(operationPath, pathItemObject);
+			paths.addPathItem(operationPath, pathItemObject);
 		}
 	}
 
--- a/springdoc-openapi-starter-webmvc-api/src/main/java/org/springdoc/webmvc/api/OpenApiResource.java
+++ b/springdoc-openapi-starter-webmvc-api/src/main/java/org/springdoc/webmvc/api/OpenApiResource.java
@@ -167,6 +167,11 @@ public abstract class OpenApiResource extends AbstractOpenApiResource {
 			Optional<ActuatorProvider> actuatorProviderOptional = springDocProviders.getActuatorProvider();
 			if (actuatorProviderOptional.isPresent() && springDocConfigProperties.isShowActuator()) {
 				Map<RequestMappingInfo, HandlerMethod> actuatorMap = actuatorProviderOptional.get().getMethods();
+				List<RequestMappingInfo> globMatchActuators = actuatorMap.keySet().stream()
+					.filter(requestMappingInfo -> requestMappingInfo.getPatternValues().stream()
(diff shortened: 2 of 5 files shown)


bnasslahsen:
Thank you @mc1arke for your contributions to the project 👍

PR #3001, opened 2025-05-21
Support for @Positive

mpleine:
Added support for @Positive annotation validation constraint

When merged, the following documentation should be updated also

    https://github.com/springdoc/springdoc.github.io/blob/master/src/docs/asciidoc/faq.adoc?plain=1#L192-L193
    https://github.com/springdoc/springdoc.github.io/blob/master/src/docs/asciidoc/intro.adoc?plain=1#L15

Suggestion:
Maybe there could be a complete list in the documentation, which validation annotations are currently supported (and the ones that are not listed there, are currently not supported)

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/utils/SchemaUtils.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/utils/SchemaUtils.java
@@ -1,8 +1,5 @@
 package org.springdoc.core.utils;
 
-import static org.springdoc.core.utils.Constants.OPENAPI_ARRAY_TYPE;
-import static org.springdoc.core.utils.Constants.OPENAPI_STRING_TYPE;
-
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.math.BigDecimal;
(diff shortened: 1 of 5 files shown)


NelsonLau88 (approved):

springdoc-openapi-starter-common/src/main/java/org/springdoc/core/utils/SchemaUtils.java
```diff
@@ -189,6 +190,9 @@ public static boolean fieldRequired(Field field, @Nullable io.swagger.v3.oas.ann
 	public static void applyValidationsToSchema(Schema<?> schema, List<Annotation> annotations) {
 		annotations.forEach(anno -> {
 			String annotationName = anno.annotationType().getSimpleName();
+			if (annotationName.equals(Positive.class.getSimpleName())) {
+				schema.setMinimum(BigDecimal.ONE);
```
kdebski85, line 194: Value `BigDecimal.ONE` is valid only for integers.
For decimals, for example, we can have a valid value `0.00001`.

kdebski85:
@bnasslahsen Any reason it was merged?
The implementation is invalid for decimal numbers. Please see my comments above.

kdebski85:
@mpleine Are you going to prepare another PR  that would fix it for decimals?

kdebski85:
I can see it was fixed in https://github.com/springdoc/springdoc-openapi/commit/6c3aa8f771ff68b8f6c1fa458c96aa18a5c5cf2f

PR #3002, opened 2025-05-21
Bump swagger-api to 2.2.32 and fix tests (wip)

lc-nyovchev:
Relates to: ISSUE https://github.com/springdoc/springdoc-openapi/issues/3003


springdoc-openapi-starter-webmvc-api/src/test/resources/results/3.1.0/app213.json
```diff
@@ -47,7 +55,7 @@
             }
           }
         },
-        "x-amazon-apigateway-integration": {
+        "x-x-amazon-apigateway-integration": {
```
lc-nyovchev, line 58: That doesn't seem correct, but that's what we get as a response

NelsonLau88 (approved):

lc-nyovchev:
Will reopen later when swagger-core is updated, for now swagger is behaving weirdly.

PR #3007, opened 2025-05-23
Fixes for Spring Boot 3.5.0 API

### LINKED ISSUE #3005: SB 3.5.x compatibility
SB 3.5.0 comes with a breaking change, a method was renamed, which causes method not found at start-up:

```
***************************
APPLICATION FAILED TO START
***************************

Description:

An attempt was made to call a method that does not exist. The attempt was made from the following location:

    org.springdoc.core.providers.HateoasHalProvider.isHalEnabled(HateoasHalProvider.java:81)

The following method did not exist:

    'boolean org.springframework.boot.autoconfigure.hateoas.HateoasProperties.getUseHalAsDefaultJsonMediaType()'

The calling method's class, org.springdoc.core.providers.HateoasHalProvider, was loaded from the following location:

    jar:file:/Users/silviu.burcea/.m2/repository/org/springdoc/springdoc-openapi-starter-common/2.8.8/springdoc-openapi-starter-common-2.8.8.jar!/org/springdoc/core/providers/HateoasHalProvider.class

The called method's class, org.springframework.boot.autoconfigure.hateoas.HateoasProperties, is available from the following locations:

    jar:file:/Users/silviu.burcea/.m2/repository/org/springframework/boot/spring-boot-autoconfigure/3.5.0/spring-boot-autoconfigure-3.5.0.jar!/org/springframework/boot/autoconfigure/hateoas/HateoasProperties.class

The called method's class hierarchy was loaded from the following locations:

    org.springframework.boot.autoconfigure.hateoas.HateoasProperties: file:/Users/silviu.burcea/.m2/repository/org/springframework/boot/spring-boot-autoconfigure/3.5.0/spring-boot-autoconfigure-3.5.0.jar
```

`getUseHalAsDefaultJsonMediaType` is now called `isUseHalAsDefaultJsonMediaType`. I have reported the issue to Spring team as well, see https://github.com/spring-projects/spring-boot/issues/45648, I'm not sure if they will revert the change or if you need to bump minor version to have SB 3.5.x compatibility. I'm happy to contribute once it's clear which project needs to change.

mschout:
Spring Boot renamed the property methods to determine if HAL is enabled or not.  Use reflection to determine which method to call so we can support both versions

Another alternative would have been to merely bump the minimum supported version of spring boot to 3.5.0 and just use the new methods directly.  Not sure what is preferred here.  This does get it to start under 3.5 and 3.4 however.

Fixes #3005

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/providers/HateoasHalProvider.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/providers/HateoasHalProvider.java
@@ -26,12 +26,15 @@
 
 package org.springdoc.core.providers;
 
+import java.util.List;
 import java.util.Optional;
 
 import jakarta.annotation.PostConstruct;
 
 import org.springframework.boot.autoconfigure.hateoas.HateoasProperties;
 import org.springframework.hateoas.mediatype.hal.Jackson2HalModule;
+import org.springframework.lang.NonNull;
+import org.springframework.util.ReflectionUtils;
 
 /**
  * The type Hateoas hal provider.
@@ -79,8 +82,26 @@ public class HateoasHalProvider {
 	 */
 	public boolean isHalEnabled() {
 		return hateoasPropertiesOptional
-				.map(HateoasProperties::getUseHalAsDefaultJsonMediaType)
+				.map(this::isHalEnabled)
 				.orElse(true);
 	}
 
+	private boolean isHalEnabled(@NonNull HateoasProperties hateoasProperties) {
+		// In spring-boot 3.5, the method name was changed from getUseHalAsDefaultJsonMediaType to isUseHalAsDefaultJsonMediaType
+		var possibleMethodNames = List.of("isUseHalAsDefaultJsonMediaType", "getUseHalAsDefaultJsonMediaType");
+
+		for (var methodName : possibleMethodNames) {
+			var method = ReflectionUtils.findMethod(HateoasProperties.class, methodName);
+			if (method != null) {
+				var result = ReflectionUtils.invokeMethod(method, hateoasProperties);
+				if (result instanceof Boolean) {
+					return (boolean) result;
+				}
+
+				throw new IllegalStateException("Method " + methodName + " did not return a boolean value");
+			}
+		}
+
+		throw new IllegalStateException("No suitable method found to determine if HAL is enabled");
+	}
 }
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/providers/HateoasHalProvider.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/providers/HateoasHalProvider.java
@@ -86,7 +86,7 @@ public class HateoasHalProvider {
 				.orElse(true);
 	}
 
-	private boolean isHalEnabled(@NonNull HateoasProperties hateoasProperties) {
+	private static boolean isHalEnabled(@NonNull HateoasProperties hateoasProperties) {
 		// In spring-boot 3.5, the method name was changed from getUseHalAsDefaultJsonMediaType to isUseHalAsDefaultJsonMediaType
 		var possibleMethodNames = List.of("isUseHalAsDefaultJsonMediaType", "getUseHalAsDefaultJsonMediaType");
 
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/providers/HateoasHalProvider.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/providers/HateoasHalProvider.java
@@ -82,7 +82,7 @@ public class HateoasHalProvider {
 	 */
 	public boolean isHalEnabled() {
 		return hateoasPropertiesOptional
-				.map(this::isHalEnabled)
+				.map(HateoasHalProvider ::isHalEnabled)
 				.orElse(true);
 	}
 
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/providers/HateoasHalProvider.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/providers/HateoasHalProvider.java
@@ -82,7 +82,7 @@ public class HateoasHalProvider {
 	 */
 	public boolean isHalEnabled() {
 		return hateoasPropertiesOptional
-				.map(HateoasHalProvider ::isHalEnabled)
+				.map(HateoasHalProvider::isHalEnabled)
 				.orElse(true);
 	}
 
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/providers/HateoasHalProvider.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/providers/HateoasHalProvider.java
@@ -94,8 +94,8 @@ public class HateoasHalProvider {
 			var method = ReflectionUtils.findMethod(HateoasProperties.class, methodName);
 			if (method != null) {
 				var result = ReflectionUtils.invokeMethod(method, hateoasProperties);
-				if (result instanceof Boolean) {
-					return (boolean) result;
+				if (result instanceof Boolean halEnabled) {
+					return halEnabled;
 				}
 
 				throw new IllegalStateException("Method " + methodName + " did not return a boolean value");


mschout:
Side note, the Link schema doesn't get registered under 3.5.0.  I did not have time to look at that yet (the test fails). I think that is unrelated to this patch.

mschout:
Yeah its this patch.  Fix incoming sorry.

mschout:
sorry for confusion.  current version of the MR works for me under both 3.4.5 and 3.5.0.  This should be good to go now.

rcbandit111:
I'm facing this issue when I tested the latest Spring Cloud version 2025.0.0. When I can expect official release with this fix?

springdoc-openapi-starter-common/src/main/java/org/springdoc/core/providers/HateoasHalProvider.java
```diff
@@ -79,8 +82,26 @@ protected void init() {
...
+				var result = ReflectionUtils.invokeMethod(method, hateoasProperties);
+				if (result instanceof Boolean) {
+					return (boolean) result;
+				}
+
+				throw new IllegalStateException("Method " + methodName + " did not return a boolean value");
+			}
+		}
+
+		throw new IllegalStateException("No suitable method found to determine if HAL is enabled");
```
StefanSchrass, line 105: Initial implementation defaults to `true` if no HateoasProperties are present. Could this also be applied here?

springdoc-openapi-starter-common/src/main/java/org/springdoc/core/providers/HateoasHalProvider.java
```diff
@@ -79,8 +82,26 @@ protected void init() {
...
+
+		for (var methodName : possibleMethodNames) {
+			var method = ReflectionUtils.findMethod(HateoasProperties.class, methodName);
+			if (method != null) {
+				var result = ReflectionUtils.invokeMethod(method, hateoasProperties);
+				if (result instanceof Boolean) {
+					return (boolean) result;
+				}
+
+				throw new IllegalStateException("Method " + methodName + " did not return a boolean value");
```
StefanSchrass, line 101: Initial implementation defaults to `true` if no HateoasProperties are present. Could this also be applied here?

springdoc-openapi-starter-common/src/main/java/org/springdoc/core/providers/HateoasHalProvider.java
kimberninger: This is already the case. Have a look at line 86 (formerly line 83):
```java
.orElse(true)
```

StefanSchrass: No its not. Line 82 did map, this line throws--bypassing--the default.

I totally get the difference here, but my point is that it is actually harmful because even if you dont use HAL at all my application wont start (and thats why I am here)

mschout: The throw on line 105 should only happen if `HateoasProperties` exists (IE the `Optional` is non-empty), but we couldn't figure out the appropriate method name to call on it.  We try the new method name in SB 3.5.0 first, and if we don't find that, we try the one from SB 3.4.x and ealier.   @StefanSchrass do you have a small working demo that does not start with the patch?

springdoc-openapi-starter-common/src/main/java/org/springdoc/core/providers/HateoasHalProvider.java
mschout: `.orElse(true)` will *still* be called if the `Optional<HateoasProperties>` is empty

StefanSchrass: totally, but its sheer existens doesn't mean that it is important to the application.

springdoc-openapi-starter-common/src/main/java/org/springdoc/core/providers/HateoasHalProvider.java
```diff
@@ -79,8 +82,26 @@ protected void init() {
...
+	private boolean isHalEnabled(@NonNull HateoasProperties hateoasProperties) {
+		// In spring-boot 3.5, the method name was changed from getUseHalAsDefaultJsonMediaType to isUseHalAsDefaultJsonMediaType
+		var possibleMethodNames = List.of("isUseHalAsDefaultJsonMediaType", "getUseHalAsDefaultJsonMediaType");
+
+		for (var methodName : possibleMethodNames) {
+			var method = ReflectionUtils.findMethod(HateoasProperties.class, methodName);
+			if (method != null) {
+				var result = ReflectionUtils.invokeMethod(method, hateoasProperties);
+				if (result instanceof Boolean) {
+					return (boolean) result;
```
Sineaggi, line 98: Should be possible to avoid the type-casting here as the project depends on java 17
```suggestion
				if (result instanceof Boolean halEnabled) {
					return halEnabled;
```

springdoc-openapi-starter-common/src/main/java/org/springdoc/core/providers/HateoasHalProvider.java
```diff
@@ -79,8 +82,26 @@ protected void init() {
 	 */
 	public boolean isHalEnabled() {
 		return hateoasPropertiesOptional
-				.map(HateoasProperties::getUseHalAsDefaultJsonMediaType)
+				.map(this::isHalEnabled)
 				.orElse(true);
 	}
 
+	private boolean isHalEnabled(@NonNull HateoasProperties hateoasProperties) {
```
kdebski85, line 89: This method can be static.

springdoc-openapi-starter-common/src/main/java/org/springdoc/core/providers/HateoasHalProvider.java
mschout: pushed.

springdoc-openapi-starter-common/src/main/java/org/springdoc/core/providers/HateoasHalProvider.java
mschout: I pushed this thanks.

silviuburceadev (approved):

sebobr (approved):

samuelstein:
still needs to be reviewed by @Sineaggi and @StefanSchrass.
btw i don't think it makes sense to have so many reviewers because it slows the process down very much.
we are waiting already two weeks for this fix

springdoc-openapi-starter-common/src/main/java/org/springdoc/core/providers/HateoasHalProvider.java
EverettZ: What's the conclusion here? @mschout added compatibility with SB 3.5.*. Great.

I wonder if maintaining the current public getter and then falling back to reflection is better? Or maybe just pin the min SB version.

bdbogjoe:
thanks, waiting for this fix to upgrade my apps  to spring boot 3.5.0

Sineaggi (approved):

kdebski85:
This PR was approved by 6 people.
Do we really need to wait for @StefanSchrass approval?

kdebski85:
This PR was merged. Thank you!

However, I can see that `main` branch already has dependency on Spring Boot 3.5:

https://github.com/springdoc/springdoc-openapi/commit/6111073e41bb61bfbc76f580097fd120ff17b154#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R14

@mschout 
Maybe we can remove SB 3.4 compatibility from `HateoasHalProvider` and use the accessor directly with `.map(HateoasProperties::isUseHalAsDefaultJsonMediaType)`. What do you think?

mschout:
@kdebski85 If I understand correctly (and I tested this on an internal project), it is still possible to use spring boot 3.4.x and springdoc-openapi 2.8.9 and the resolved version of SB will be 3.4.x.

E.g.:

```
plugins {
  id("org.springframework.boot") version "3.4.6"
}

...
dependencies {
  implementation("org.springdoc:springdoc-openapi-starter-webmvc-ui:2.8.9")
}
```

then:

```
./gradlew dependencyInsight --dependency org.springframework.boot:spring-boot-starter
```

And it picks (correctly) the 3.4.6 version of SB.


```
org.springframework.boot:spring-boot-starter:3.4.6 (selected by rule)
```

So given that, and 3.5 is still on a `.0` release maybe better to hold off for a bit before dropping the reflection IMO, but if the team feels differently here, that's fine as well.

silviuburceadev:
I've tested the fix with 3.5.0 and it works just fine :)

PR #2600, opened 2024-05-24
Fix consumes and produces calculation. Fixes #2596

jakvbs:
fixes #2596

PR #3026, opened 2025-06-21
Fix unexpected merging of media types

### LINKED ISSUE #3023: Unexpected merging of media types
**Describe the bug**

According to the Javadoc for `org.springframework.web.bind.annotation.RequestMapping#produces`, method-level `produces` should override class-level declarations. 

> Supported at the type level as well as at the method level! If specified at both levels, the method level produces condition overrides the type level condition.

However, when both are used, the springdoc generated OpenAPI spec merges them.

**To Reproduce**

- Spring Boot 3.5.0
- springdoc-openapi-starter-webmvc-ui 2.8.9

Example controller:
```
@RestController
@RequestMapping(produces = MediaType.APPLICATION_JSON_VALUE)
public class MediaTypeDemoController {

    @GetMapping("/json")
    public ResponseEntity<Map<String, String>> getJsonResponse() {
        Map<String, String> response = Map.of("message", "This is JSON");
        return ResponseEntity.ok(response);
    }

    @GetMapping(value = "/xml", produces = MediaType.APPLICATION_XML_VALUE)
    public ResponseEntity<String> getXmlResponse() {
        String xml = "<message>This is XML</message>";
        return ResponseEntity.ok(xml);
    }
}
```

Generated spec:
``` yaml 
{
  openapi: "3.1.0",
  info: {
    title: "OpenAPI definition",
    version: "v0"
  },
  servers: [
    {
      url: "http://localhost:8080/",
      description: "Generated server url"
    }
  ],
  paths: {
    /xml: {
      get: {
        tags: [
          "media-type-demo-controller"
        ],
        operationId: "getXmlResponse",
        responses: {
          200: {
            description: "OK",
            content: {
              application/xml: {
                schema: {
                  type: "string"
                }
              },
              application/json: {
                schema: {
                  type: "string"
                }
              }
            }
          }
        }
      }
    },
    /json: {
      get: {
        tags: [
          "media-type-demo-controller"
        ],
        operationId: "getJsonResponse",
        responses: {
          200: {
            description: "OK",
            content: {
              application/json: {
                schema: {
                  type: "object",
                  additionalProperties: {
                    type: "string"
                  }
                }
              }
            }
          }
        }
      }
    }
  },
  components: { }
}
```

**Expected behavior**

- The OpenAPI spec generated for `GET /xml` includes both `application/json` and `application/xml` as possible response media types. I expected it to only list `application/xml`, since it's explicitly declared at the method level.

**Additional context**
- Behavior was probably introduced by #2600

Mattias-Sehlstedt:
Fixes #3023

A minimal suggested fix. Given the description provided in the issue, I do not see how [mergeArrays](https://github.com/springdoc/springdoc-openapi/pull/2600/files#diff-60dbe5350c24bb6f53b8fd8216bd7668923262bf47d2f61a3317774d6b6d6a8aR308-R311) should ever be involved with any media type processing, since the media type documentation states that they should override rather than join the types. But due to the lack of descriptive testing together with values being set upon class instantiation and then later changed with a method invocation, I am hesitant to do too much of a change.

Feel free to use and contribute with the unit tests in this PR if you implement a change, since they add a lot of information with regard to what is an implicit/explicit behavior.

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/models/MethodAttributes.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/models/MethodAttributes.java
@@ -270,7 +270,11 @@ public class MethodAttributes {
 		RequestMapping reqMappingClass = AnnotatedElementUtils.findMergedAnnotation(method.getDeclaringClass(), RequestMapping.class);
 
 		if (reqMappingMethod != null && reqMappingClass != null) {
-			fillMethods(ArrayUtils.addAll(reqMappingMethod.produces(), reqMappingClass.produces()), ArrayUtils.addAll(reqMappingMethod.consumes(), reqMappingClass.consumes()), reqMappingMethod.headers());
+			fillMethods(
+					calculateMethodMediaTypes(reqMappingMethod.produces(), reqMappingClass.produces()),
+					calculateMethodMediaTypes(reqMappingMethod.consumes(), reqMappingClass.consumes()),
+					reqMappingMethod.headers()
+			);
 		}
 		else if (reqMappingMethod != null) {
 			fillMethods(reqMappingMethod.produces(), reqMappingMethod.consumes(), reqMappingMethod.headers());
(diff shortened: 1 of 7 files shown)


bnasslahsen:
@Mattias-Sehlstedt,

Thank you so much for your contribution.

PR #3038, opened 2025-07-06
Fixed couple of tests and bumped swagger-core version

lc-nyovchev:
This is DRAFT cause there are some regressions on a couple of tests, it seems.


bnasslahsen:
see c7fe2b2023a65848a464948239c9228078f5095e

Issue #3056, opened 2025-08-12
Customizer for @Webhooks


zdary:
**Is your feature request related to a problem? Please describe.**
@Webhooks are included in OpenAPI automatically without a possibility to exclude them from a group.

When multiple GroupedOpenApi are defined all Webhooks are included in all groups. 
Some Webhooks are relevant for certain groups only. 

**Describe the solution you'd like**

Allow customization over the Webhook inclusion.
i.e.
```
GroupedOpenApi.builder()
  .addOpenApiCustomizer(openApi -> {
    openApi.addWebhooksFromClass(MyWebhookDefinitions.class);
  })
```
 
the invocation of `calculateWebhooks(openAPI, locale);` method in `AbstractOpenApiResource.getOpenApi` method should be wrapped
```
if (springDocConfigProperties.isWebhookDetectionEnabled()) {
  calculateWebhooks(openAPI, locale);
}
```
**Describe alternatives you've considered**

I've tried to get into `AbstractOpenApiResource.calculateWebhooks`. That method is protected so customization is possible but not a clean solution. 

thank you

zdary:
Solution proposal for code review: https://github.com/springdoc/springdoc-openapi/pull/3057

PR #3057, opened 2025-08-12
Refactor webhook discovery and scanning mechanism

zdary:
**Implement package-based filtering for webhooks**
This commit introduces support for filtering webhooks based on the `springdoc.packages-to-scan` and `springdoc.packages-to-exclude` properties. This provides users with more granular control over which webhooks are included in the generated OpenAPI specification, aligning their behavior with that of controllers.

To verify and illustrate this new functionality, a new application test (`SpringDocApp246Test`) has been added. This test:
•Defines webhook components in two separate packages.
•Configures one package to be included and the other to be excluded.
•Asserts that the final OpenAPI document correctly contains only the webhooks from the scanned package.

Additionally, a new abstract base class, `AbstractSpringDocV31Test`, has been created to streamline the setup for OpenAPI 3.1 integration tests.

Fixes https://github.com/springdoc/springdoc-openapi/issues/3056

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java
@@ -140,6 +140,7 @@ import static org.springframework.util.AntPathMatcher.DEFAULT_PATH_SEPARATOR;
  * @author kevinraddatz
  * @author hyeonisism
  * @author doljae
+ * @author zdary
  */
 public abstract class AbstractOpenApiResource extends SpecFilter {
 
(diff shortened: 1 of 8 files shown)


AlexanderBartash (approved):

Mattias-Sehlstedt:
Hi @zdary

I would suggested adding an application test to further illustrate the changed behavior. So something similar to [this test](https://github.com/springdoc/springdoc-openapi/blob/main/springdoc-openapi-starter-webmvc-api/src/test/java/test/org/springdoc/api/v31/app9/WebHookResource.java), but that it explicitly covers how the new filtering works (the test produces and exposes a specification for the code, and the expected specification is then located in resources/results).

I am not affiliate with this project in any way, so only a suggestion rather than a review.

zdary:
> Hi @zdary
> 
> I would suggested adding an application test to further illustrate the changed behavior. So something similar to [this test](https://github.com/springdoc/springdoc-openapi/blob/main/springdoc-openapi-starter-webmvc-api/src/test/java/test/org/springdoc/api/v31/app9/WebHookResource.java), but that it explicitly covers how the new filtering works (the test produces and exposes a specification for the code, and the expected specification is then located in resources/results).
> 
> I am not affiliate with this project in any way, so only a suggestion rather than a review.

good point, I have added the tests to cover the filtering. thank you

PR #3092, opened 2025-09-16
feat: Isolate Core interfaces

alessiomatricardi:
Isolate commonly used interfaces in a separate module `springdoc-openapi-common`.
The aim of this PR is to make core interfaces/annotations usable in other libraries without any Spring Boot autoconfiguration.


bnasslahsen:
@alessiomatricardi,

Thanks for your contribution!
This project is primarily intended for use with Spring Boot.
For projects that don’t rely on the Spring Boot base dependencies, there’s limited value in integrating it — especially if the goal is to use it alongside springdoc-openapi.

alessiomatricardi:
@bnasslahsen 
My goal is not to use this library without Spring Boot, I will make an example to be more clear.
Working on a modularized library, It is useful to have some classes like ParameterObject available during interfaces definition, and have the complete springdoc-openapi (with all its boot definitions) only in the "starter" module.
It was intended to be only a concern separation between descriptive interfaces and bare implementation.
For those who was using the library as is, nothing will have changed.
But understand your point, thank you 😃

PR #3105, opened 2025-10-10
Upgrade swagger-core to version 2.2.38

Mattias-Sehlstedt:
Upgrades swagger-core from 2.2.36 to 2.2.38. 

Note that the schema type overriding for the 3.1 specification resolver seems to have had some minor regression, as it now seems to require the 3.1 `types` to be set rather than the old 3.0 `type`. This has previously been highlighted as issues by consumers, so might be necessary to investigate if a fallback is possible.

This should also close https://github.com/swagger-api/swagger-core/issues/4822, since the duplicated `oneOf` issue seems to have been resolved.


bnasslahsen:
@Mattias-Sehlstedt,

Thanks a lot for your great contributions to the project!

PR #3123, opened 2025-10-31
feat: static resources

TiPo4u4eK:
Support for static resources

### DIFF
--- a/springdoc-openapi-starter-webflux-api/src/main/java/org/springdoc/webflux/core/visitor/RouterFunctionVisitor.java
+++ b/springdoc-openapi-starter-webflux-api/src/main/java/org/springdoc/webflux/core/visitor/RouterFunctionVisitor.java
@@ -26,19 +26,24 @@
 
 package org.springdoc.webflux.core.visitor;
 
+import java.lang.reflect.Field;
 import java.util.ArrayList;
+import java.util.Set;
 import java.util.function.Function;
 
 import org.springdoc.core.fn.AbstractRouterFunctionVisitor;
 import reactor.core.publisher.Mono;
 
 import org.springframework.core.io.Resource;
+import org.springframework.http.HttpMethod;
 import org.springframework.web.reactive.function.server.HandlerFunction;
 import org.springframework.web.reactive.function.server.RequestPredicate;
 import org.springframework.web.reactive.function.server.RequestPredicates;
 import org.springframework.web.reactive.function.server.RouterFunction;
 import org.springframework.web.reactive.function.server.RouterFunctions;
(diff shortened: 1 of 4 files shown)


bnasslahsen:
@TiPo4u4eK,

Could you please provide a more detailed description of the feature you’ve added?
Additionally, the MVC implementation is required for feature parity when it;'s possible—please include it as well.

TiPo4u4eK:
Sure, @bnasslahsen 

This feature enables support for serving resources, making it possible to expose them through OpenAPI documentation. In the current SpringDoc implementation, using RouterFunction and RouterOperation for routing and documenting endpoints requires special handling. Without necessary adjustments to RouterFunctionVisitor, attempting to access the OpenAPI documentation would fail. This is required feature because many organizations enforce the explicit use of RouterOperation when defining routes with RouterFunction.

For example, I used ClassPathResource to serve icons, which then process with ResourcesRouterFunction for handling of resources.

I don't think MVC requires any updates. Resources are commonly served through the ResourceHandlerRegistry in a configuration class, like so:

`
@Configuration
public class WebConfig implements WebMvcConfigurer {
    @Override
    public void addResourceHandlers(ResourceHandlerRegistry registry) {
        registry.addResourceHandler("/icons/**")
                .addResourceLocations("classpath:/icons/")
                .setCachePeriod(3600);
    }
}
`

To include such resource handling in documentation, the most common approach is to use a custom io.swagger.v3.oas.models.OpenAPI. There may be other ways, but this is the most common way I've seen to provide resources in OpenAPI documentation.

TiPo4u4eK:
Sure, @bnasslahsen 

This feature enables support for serving resources, making it possible to expose them through OpenAPI documentation. In the current SpringDoc implementation, using RouterFunction and RouterOperation for routing and documenting endpoints requires special handling. Without necessary adjustments to RouterFunctionVisitor, attempting to access the OpenAPI documentation would fail. This is required feature because many organizations enforce the explicit use of RouterOperation when defining routes with RouterFunction.

For example, I used ClassPathResource to serve icons, which then process with ResourcesRouterFunction for handling of resources.

I don't think MVC requires any updates. Resources are commonly served through the ResourceHandlerRegistry in a configuration class, like so:

```
@Configuration
public class WebConfig implements WebMvcConfigurer {
    @Override
    public void addResourceHandlers(ResourceHandlerRegistry registry) {
        registry.addResourceHandler("/icons/**")
                .addResourceLocations("classpath:/icons/")
                .setCachePeriod(3600);
    }
}
```

To include such resource handling in documentation, the most common approach is to use a custom io.swagger.v3.oas.models.OpenAPI. There may be other ways, but this is the most common way I've seen to provide resources in OpenAPI documentation.

Issue #3116, opened 2025-10-25
[Feature Request] Automatically append Enum descriptions from Request DTOs to Operation Description


TAEWOOKK:
<img width="1144" height="671" alt="Image" src="https://github.com/user-attachments/assets/3d646b88-3925-49a7-a457-8b26ec7468ad" />

Is your feature request related to a problem? Please describe. Yes. When our API Request DTOs contain enum fields, our front-end developers or API consumers have to click the "Schema" tab in Swagger UI just to see what the valid string values (keys) are and what they mean.

This is inconvenient. The alternative is to manually write all enum details (e.g., "ACTIVE: Means the user is active...") into the @Operation(description = "...") annotation for every API.

This manual process is tedious, error-prone, and violates the DRY (Don't Repeat Yourself) principle. If an enum is updated (e.g., a new value is added), developers must remember to manually update the description string, which is often forgotten.

What is the actual result using OpenAPI Description (yml or json)? The operation.description is empty or only contains text that was manually entered. It does not automatically reflect the details of enum fields used in the request.

Describe the solution you'd like I would like springdoc-openapi to provide a built-in feature or a standard OpenApiCustomizer that automatically scans the parameters of an operation (especially @RequestBody DTOs).

If it finds enum fields, it should:

Find all constants of that enum.

Conventionally look for a specific method on the enum constant (e.g., getDescription(), getValue(), or similar) to get a human-readable description.

Automatically append a formatted (e.g., Markdown) list of the enum constants and their descriptions to the operation.description.

What is the expected result using OpenAPI Description (yml or json)? The operation.description field should be automatically enriched. For example:

(Existing manual description...)

Markdown

---
**✅ `status` Enum Details**
- `ACTIVE`: Represents an active user.
- `INACTIVE`: Represents an inactive user.
- `PENDING`: Represents a user pending activation.
Describe alternatives you've considered

Manual Annotation: Manually writing this information in @Operation or @Schema for every single API. This is the current, inefficient workaround.

Per-Project Customizer: Using a custom OpenApiCustomizer in every project. It would be much better if this was a built-in, configurable feature of springdoc-openapi itself.

Additional context I have already implemented a proof-of-concept OpenApiCustomizer that achieves this. If this feature is desirable and you are open to contributions, I would be happy to open a Pull Request with the implementation.

bnasslahsen:
@TAEWOOKK,

Are you willing to propose a PR for this enhancement request ?

[...]

PR #3125, opened 2025-11-01
feat: Add @EnumDescription annotation for automatic enum

TAEWOOKK:
## Description

This PR implements automatic enum description injection into OpenAPI operation descriptions. 
It addresses the issue described in [#3116](https://github.com/springdoc/springdoc-openapi/issues/3116)

### What's Changed

- Added `@EnumDescription` annotation to mark enum fields for automatic description generation
- Automatically extracts enum constants and their descriptions from annotated fields
- Formats enum information as Markdown and appends to operation descriptions

### Usage Example

```java
public enum UserStatus {
    ACTIVE("Represents an active user"),
    INACTIVE("Represents an inactive user"),
    PENDING("Represents a user pending activation");

    private final String description;

    UserStatus(String description) {
        this.description = description;
    }

    public String getDescription() {
        return description;
    }
}

public class UserRequest {
    @EnumDescription  // or @EnumDescription(fieldName = "label")
    private UserStatus status;
    
    private String name;
}
```

image
image

### Technical Details

- **Customizer**: `EnumDescriptionOperationCustomizer` implements `GlobalOperationCustomizer`
- **Bean Registration**: Automatically registered in `SpringDocConfiguration`
- **Field Scanning**: Recursively scans DTO classes including parent classes
- **Reflection**: Uses reflection to read enum field values safely

### Future Improvements

As mentioned in the original issue, we considered adding support for schema-level descriptions as well. 
This could be implemented using `PropertyCustomizer` to inject enum descriptions into schema property descriptions. 
This enhancement could be added in a future PR if there's interest.


Mattias-Sehlstedt:
I would suggest adding a test case for when the annotation is used. This both as an illustrative case for how to use it, but also to show that the current implementation actually behaves as expected when utilized. Use [the current tests](https://github.com/springdoc/springdoc-openapi/tree/main/springdoc-openapi-starter-webmvc-ui/src/test/java/test/org/springdoc/ui) as a suggestion for how they usually are structured.

TAEWOOKK:
Thanks for the feedback!
I'd like to clarify where to add the test case. I notice there are two different test patterns in the project:

1. `springdoc-openapi-starter-common/src/test/java/org/springdoc/core/` 
   - Unit tests using `WebApplicationContextRunner` (like `SpringDocHateoasConfigurationTest`)

2. `springdoc-openapi-starter-webmvc-api/src/test/java/test/org/springdoc/api/v30/appXXX/`
   - Integration tests following the `appXXX` pattern

Note that the `EnumDescription` annotation and `EnumDescriptionOperationCustomizer` are both implemented in the `common` module, which is why I'm unsure which test pattern to follow.

Which directory structure should I use for the `@EnumDescription` test case?

Mattias-Sehlstedt:
I prefer the second approach with an integration test as per `appXXX`. So one can easily see exactly how the generated OpenAPI specification looks.

TAEWOOKK:
This PR adds test cases to verify the EnumDescription annotation functionality:

- Test for default `description` field
- Test for custom `fieldName` (e.g., `label`)

The tests verify that enum information is properly added to OpenAPI operation descriptions.

Please review when convenient. Thanks!

Mattias-Sehlstedt:
I see now that there exists two different ways of doing integration tests, either in the `x-api` or the `x-ui` folder. Could we replicate your test so that it is also present in as a [x-api](https://github.com/springdoc/springdoc-openapi/tree/main/springdoc-openapi-starter-webmvc-api/src/test/java/test/org/springdoc/api/v30) test?

The difference is that the assertion is made against the entire `apixxx.json`, so it is easy to see exactly how the exposed specification looks like.

---
A bit off topic, but what is the reason for deriving the name from a potential enum value, rather than explicitly having the values in a `@Schema` annotation on the class that carries the information? Or are we sending a domain object out directly with the api? I always try to segment the domain and the api with pure DTO objects, that are responsible for de(serialization) and carrying any meaningful information to the client through the specification (so `@Schema` annotation).

So the domain is:
```java
public enum UserStatus {
    ACTIVE,
    INACTIVE,
    PENDING;
}
```
while the DTO is
```java
@Schema(name = "UserStatus", description = "...", example = "...", ....)
public enum UserStatusDTO {
    ACTIVE,
    INACTIVE,
    PENDING;
}
```
The current approach seems like it could potentially cause minor "leakage", since it is not intuitive that anything that goes into an enum's value could reach the client. Sure it is semi-safe, since it is opt-in, but I would always argue for a more strict domain/api segmentation.

TAEWOOKK:
Thank you for the detailed feedback. I completely agree with your point about the importance of strict Domain-API segmentation.

My primary motivation for implementing @EnumDescription was to solve a clear pain point for frontend developers. They often find it difficult to check the available values and descriptions for an Enum used in API requests/responses. This feature automates embedding this information directly into the operation's description, allowing them to understand and use the API much faster.

I fully understand your concern about potential information 'leakage'. As you rightly pointed out, the ideal and safest approach is to separate the internal Domain Enum (like UserStatus) from the DTO Enum (like UserStatusDTO) used for API specifications.

This feature was designed with exactly that pattern in mind. The intention is that developers will use @EnumDescription on the DTO Enum field, not the internal Domain Enum.

While this still "exposes" information, it's no longer an uncontrolled leak, but rather a controlled specification. The developer explicitly decides which Enum (the DTO-safe one) to expose. I believe the benefit of providing clear, in-spec documentation to frontend developers outweighs the risk of this controlled exposure.

This is also precisely why I added the methodName (or fieldName) option—to provide granular control. An Enum can have multiple methods:

An internal-logic method (e.g., getInternalReason())

A public-facing description method (e.g., getLabel() or getDescription())

By allowing developers to specify @EnumDescription(methodName = "getLabel"), they can explicitly choose the "safe" method for documentation. This mechanism directly prevents the accidental leakage of internal data and ensures only public-facing information is used.

In summary, this feature doesn't prevent or discourage good architecture (like Domain/DTO separation); it enhances it by bridging the information gap between a well-defined DTO (Enum) and the frontend developer who needs to consume it.

Thank you again for the insightful feedback, and I look forward to hearing your thoughts.

Mattias-Sehlstedt:
Thanks for the additional test and the extensive explanation regarding the domain architecture.

I find that this looks to be a very helpful approach for the case that you are describing, and since it is opt-in it is totally up to the user to decide what data they expose.

Since you have built a good description concatenation, I think it could be nice to showcase that in some test too. I have created some examples [here](https://github.com/springdoc/springdoc-openapi/compare/main...Mattias-Sehlstedt:springdoc-openapi:TAEWOOKK-additional-examples#diff-39e60efe4353feec20c0254cc6be5f6b64113adf3a1e96d753e137ea1e190913) while I investigated some other behavior too.

For the `Field Scanning: Recursively scans DTO classes including parent classes` feature listing, is this something that has been dropped? I tried to create a deep model, but it is not processed. And when I look into the code, we only scan the operation and its direct properties.

---------------------------------------------------------------------------------------------------------------

Going off track again. The more I think about the usage the more I am curious of whether we could aid the frontend developers even more. My understanding as of now is that the frontend developers look at the swagger presentation of the specification when they build a client integration towards it? So the for example want to define the DTO/the request that should be sent towards an endpoint, and for it they will of course want to mirror the valid values enum parameters. This then would help them define the valid value space for enum field close to the operation (request body, query parameter...), since they would see it within the swagger presentation. 

Have mimicking or using tools like https://github.com/OpenAPITools/openapi-generator been considered? What they do is that they can generate client DTOs based upon a specification. I personally use the generator in some projects, and it together with `io.swagger.v3.core.jackson.ModelResolver.enumsAsRef = true;` within the server application makes enums extremely easy to view and use.

TAEWOOKK:
Thank you for the detailed feedback and the excellent test case suggestions! I've incorporated them and they really helped solidify the feature. example [here](https://github.com/springdoc/springdoc-openapi/pull/3125/commits/952d3a8814fc8bd1107c8b9eabd426993443ce08) !!!

I'd like to clarify a couple of points regarding my motivation and the feature's intent:

1. Clarification on "Recursive Parent Scan"
My apologies for the ambiguous term. When I mentioned "recursive parent scan," I was referring to scanning parent classes via inheritance (extends), not scanning nested objects.

The test5 case is designed to demonstrate this exact behavior:
TestRequest5 extends BaseRequest.
BaseRequest has an @EnumDescription on its status field.
TestRequest5 has its own @EnumDescription on its priority field.
The result correctly combines the descriptions from both the parent class and the child class.

I absolutely agree that for a frontend developer's primary integration task, using OpenAPI Generator + enumsAsRef = true can be an extremely helpful, and often ideal, workflow for handling enums.

My goal with @EnumDescription was to provide a complementary benefit. I believe that having immediate and visible enum validation directly in the Swagger UI itself also significantly helps a developer's initial understanding of the API.

This was also the main reason for adding the fieldName option. It allows developers to specify human-readable descriptions for each enum value (e.g., from a getLabel() method) and display them clearly in the UI. This helps all developers (not just frontend, but also new backend team members or QA) understand the API's business logic more quickly.

Additionally, I was personally motivated by the inconvenience of the existing alternative. To show enum values in the past, I often had to create many @ExampleObject annotations, like this:

```
@Operation(
    requestBody = @RequestBody(
        content = @Content(
            examples = {
                @ExampleObject(
                    name = "Pending Order",
                    value = "{\n" +
                            "  \"priority\": \"HIGH\",\n" +
                            "  \"status\": \"PENDING\"\n" +
                            "}"
                ),
                @ExampleObject(
                    name = "Shipping Order",
                    value = "{\n" +
                            "  \"priority\": \"URGENT\",\n" +
                            "  \"status\": \"SHIPPING\"\n" +
                            "}"
                )
            }
        )
    )
)

```
This approach made the controller code very verbose and difficult to maintain.

So, my primary purpose was to leverage the description field to provide a simple, clean way to specify this information directly in the UI—solving this maintenance problem while also improving immediate clarity for anyone reading the documentation.

Thank you for helping me clarify this!

bnasslahsen:
@Mattias-Sehlstedt,

Do we feel we are ready to merge this PR ?

Mattias-Sehlstedt:
TL;DR: Yes, but depends on what stance the project has on managing client-specific functionality (functionality that is most likely not going to be used by more than one client).

---

Yes I believe it delivers on the functionality presented in the issue, and it illustrates that it works and how with the tests provided.

The only thing I am a bit dubious about is its reusability. To me this is an intermediate support tool for aiding with understanding a specification's enums when using Swagger-UI. I would always advice to instead use structures supported by the OpenApi-specification to handle this. For example by using enums-as-refs, or by using vendor extensions to carry the descriptions. Such an approach is beneficial since it provides better support for code-generation-tools to analyze the structure and also forward this metadata to the developer in generated code.

Having the client's integration process be to use the Swagger-UI to construct objects sounds like a rare process to me, and that there are better ways to handle it "not-too-far-away". For example by having all enums be refs, since that would allow the client to look at all enum refs, implement them and their descriptions in an object, and then reuse that for all relevant endpoints. But I can also understand that one does not change their process over a day. 

So comes down to to what degree the project wants to carry client-specific helper-functionality.

bnasslahsen:
Thanks @TAEWOOKK for your contribution.

We won’t merge this PR because it introduces a specialized annotation that would increase long-term maintenance. 
Projects can reuse the code from the PR if and when they need it.

TAEWOOKK:
Thanks for the clarification. While it's unfortunate that this couldn't be merged, I respect the decision to prioritize maintainability.

Big thanks to @Mattias-Sehlstedt for the constructive discussion on domain/API separation. I'll make sure this code remains accessible for others who might need this specific feature. I look forward to contributing again in the future!

Issue #3147, opened 2025-11-19
Docs: Add Spring Boot 3 management-port + actuator exposure example for /v3/api-docs and Swagger UI


RituMalav:
### Summary
Request to add a short docs section showing how to expose `/v3/api-docs` and Swagger UI when Spring Boot 3 uses a separate **management port**.

### Problem
Many apps run actuator on a different management port. New users are unsure which properties are required so that `/v3/api-docs` and Swagger UI are available on the management port, and what URLs to hit.

### Proposal
Add a docs snippet with:
- `application.yml` showing `management.server.port`, `management.endpoints.web.exposure.include`, and `springdoc` properties.
- Expected URLs for `/actuator`, `/v3/api-docs`, and `/swagger-ui/index.html` when management port is enabled.
- One-line `curl` example to verify.

### Plan
I will open a PR adding a small “Management port + Actuator” example to the README (or the appropriate docs page) with a minimal configuration and verification steps.

RituMalav:
Opened PR #3148 to add the docs section. Happy to adjust wording/placement.

PR #3148, opened 2025-11-19
docs(management-port): add example for /v3/api-docs and Swagger UI (#…

RituMalav:
### What
Add a short README section showing how /v3/api-docs and Swagger UI behave when Spring Boot 3 uses a separate management port.

### Why
Many users put Actuator on a different management port and are unsure which URLs to use. This adds a minimal config + verification commands.

### Notes
- Content placed above “Acknowledgements”.
- References issue #3147.


RituMalav:
First-time contributor: docs-only change tied to #3147. CI shows “Failed to determine (log)”. Could you please trigger the build or label this as documentation? Happy to adjust wording or placement. Thank you!

PR #3226, opened 2026-02-24
fix: propagate JsonView context when resolving Page<T> schema

Akila-1329:
## Problem
When a controller method is annotated with @JsonView and returns Page<T>, 
the generated OpenAPI schema ignores the JsonView and exposes all fields 
of T instead of only the fields marked with that view.

## Root Cause
In PageOpenAPIConverter.resolvePagedModelType(), a new AnnotatedType is 
created for PagedModel<T> but only copies the type information, dropping 
the ctxAnnotations which carry the @JsonView context.

## Fix
Pass the original AnnotatedType to resolvePagedModelType() and copy its 
ctxAnnotations onto the new AnnotatedType, preserving the JsonView context.

## Test
Added test app245 which verifies that @JsonView(Hello.class) on a method 
returning Page<HelloWorld> correctly generates HelloWorld_Hello schema 
with only the hello field.


bnasslahsen:
Thanks @Akila-1329  for your contribution to the project!

PR #3235, opened 2026-03-02
fix: PropertyResolverUtils retains a JsonNode when reading an ExtensionProperty annotation

### LINKED ISSUE #3234: PropertyResolverUtils.resolveExtensions converts ObjectNode instances to strings
**Describe the bug**

PropertyResolverUtils.resolveExtensions will convert ObjectNode instances to Strings in cases where the extension value is a HashMap 

If value1, i.e. the hashmap value, is an ObjectNode, it will be converted to a String by the line below:

`String value1Resolved = this.resolve(value1.toString(), locale);`

A possible solution is to only call "resolve" if "value1 instanceof String", as is done for non-hashmap extension values.

**To Reproduce**
- What modules and versions of springdoc-openapi are you using? 3.0.1

We have extension properties as part/children of extension-values where the extension property is a json-object, example below:

```
  @Extension(
                                 name = "x-my-extension",
                                 properties = {
                         		   @ExtensionProperty(
                                                  name = "working-hours",
                                                  parseValue = true,
                                                  value = "{  \"working-hours\": \"Weekdays 00:15 - 22:59 CET\" }"
                                         ),
```

The output of this is the following

    "x-my-extension": {
      "working-hours": "{\"working-hours\":\"Weekdays 00:15 - 22:59 CET\"}",

I.e. the value is a string and not a JSON object.

**Expected behavior**

I expect the result to be a JSON object, like below:

    "x-my-extension": {
      "working-hours": { "working-hours": "Weekdays 00:15 - 22:59 CET" },

**Screenshots**

**Additional context**

Mattias-Sehlstedt:
Fixes #3234 as per the suggestion provided in the issue.

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/utils/PropertyResolverUtils.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/utils/PropertyResolverUtils.java
@@ -234,7 +234,12 @@ public class PropertyResolverUtils {
 					Map<String, Object> valueResolved = new HashMap<>();
 					valueAsMap.forEach((key1, value1) -> {
 						String key1Resolved = resolve(key1.toString(), locale);
-						String value1Resolved = resolve(value1.toString(), locale);
+						Object value1Resolved;
+						if (value1 instanceof String valueAsString) {
+							value1Resolved = resolve(valueAsString, locale);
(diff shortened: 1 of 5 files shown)


martinhallerdal:
Nice, thanks! :-)

PR #3245, opened 2026-03-13
Upgrade swagger-core from version 2.2.43 to 2.2.45

Mattias-Sehlstedt:
Known caveats
1. Change https://github.com/swagger-api/swagger-core/pull/5018 that introduced support for `@Nullable` leaks the nullable to array items. Reported in https://github.com/swagger-api/swagger-core/issues/5077. The annotation also incorrectly leaks to objects, so that they will be marked `nullable: true` or `type: null`. I know the cause and the fix but I am waiting for the first ticket to be resolved before I implement that fix.
2. Change https://github.com/swagger-api/swagger-core/pull/5063 made it so that the swagger-core ObjectMapper is need to properly retain the null default/example when cloning via json. This means that custom injected extensions that do not abide to having a prefix "`x-`" gets dropped.
3. I suspect change https://github.com/swagger-api/swagger-core/pull/5034 is what makes it so `"additionalProperties": {}` becomes `"additionalProperties": {"type": "object"}` This new definition does not mean the same as the first, and I would argue that the first is more correct for `Map<String, Object>` in Java. But it depends on if `Object` in the map is expected to be a Json-deserializable value or something more flat/simple.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,7 +55,7 @@
 		<central-publishing-maven-plugin.version>0.7.0
 		</central-publishing-maven-plugin.version>
 		<flatten-maven-plugin.version>1.5.0</flatten-maven-plugin.version>
-		<swagger-api.version>2.2.43</swagger-api.version>
+		<swagger-api.version>2.2.45</swagger-api.version>
 		<swagger-ui.version>5.32.0</swagger-ui.version>
 		<gmavenplus-plugin.version>1.13.1</gmavenplus-plugin.version>
 		<jjwt.version>0.9.1</jjwt.version>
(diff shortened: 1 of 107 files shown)


bnasslahsen:
Hi @Mattias-Sehlstedt,

Thank you for your efforts - i have used your initial work and comments to finish the upgrade.

Mattias-Sehlstedt:
Once swagger-core fixes the leaking `@Nullable` annotation, should I try to ensure that the local adjustments introduced here are removed? The same applies for them incorrectly classifying `additionalProperies`. I would assume that the entire  [method](https://github.com/springdoc/springdoc-openapi/commit/05957ce95d0deeb054a3ddbcddc1dd03c6f25f5f#diff-cd8ad46e4549ccf2444c512bbe5bf0ca81415fd29be0b0939ddc88b915f4c48fR193) should not even be necessary if swagger-core had the expected behavior. Same goes for the changes made [in this class](https://github.com/springdoc/springdoc-openapi/commit/05957ce95d0deeb054a3ddbcddc1dd03c6f25f5f#diff-136e27e11399f27770877095d034cd37a66a9e3faf417dc3177cd4325ac1779c).

bnasslahsen:
Agree!

PR #3249, opened 2026-03-20
fix: skip url param when urls already populated to fix BaseLayout + GroupedOpenApi break (fixes #3243)

### LINKED ISSUE #3243: swagger-ui BaseLayout config + GroupedOpenApi makes swagger-ui breaks
**Describe the bug**

Adding springdoc.swagger-ui.layout=BaseLayout in application.properties while using GroupedOpenApi group makes swagger-ui return `No API definition provided.`

Image

**To Reproduce**
using spring-boot-starter-parent v3.5.5 & springdoc-openapi-starter-webmvc-ui 2.8.12:

 -  Add `springdoc.swagger-ui.layout=BaseLayout` to application.properties.
 -  Add GroupedOpenApi beans in SwaggerConfig.java

If I remove springdoc.swagger-ui.layout=BaseLayout, it works normally:

Image

souravCoder1:
## Description

When `springdoc.swagger-ui.layout=BaseLayout` is set alongside `GroupedOpenApi` beans, swagger-ui renders "No API definition provided."

### Root cause

In `SwaggerUiConfigParameters#getConfigParameters()`, the single `url` property is unconditionally written into the config params map even when the `urls` array is already populated by group definitions. swagger-ui treats a non-null `url` as the primary source and ignores `urls`, so no API definition is found.

### Fix

Guard the `URL_PROPERTY` write behind a `CollectionUtils.isEmpty(urls)` check so it is only emitted when no grouped URLs are present. This restores correct behaviour for both single-api (url) and multi-group (urls) setups regardless of the `layout` property.

Fixes #3243

## Changes made
- `SwaggerUiConfigParameters.java`: wrap `SpringDocPropertiesUtils.put(URL_PROPERTY, ...)` in an `if (CollectionUtils.isEmpty(urls))` guard

## How has this been tested?
- Reproduced the bug locally with `springdoc.swagger-ui.layout=BaseLayout` + two `GroupedOpenApi` beans
- Confirmed swagger-ui loads correctly after the fix
- Confirmed single-API setup (no groups, layout=BaseLayout) still works — url is still emitted
- Ran `./mvnw verify` — all existing tests pass

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code so it is easy to understand
- [x] My changes generate no new warnings
- [x] No new dependencies introduced

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/properties/SwaggerUiConfigParameters.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/properties/SwaggerUiConfigParameters.java
@@ -39,6 +39,7 @@ import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.StringUtils;
 
+import org.springdoc.core.utils.SpringDocPropertiesUtils;
 import org.springframework.util.CollectionUtils;
 
 import static org.springframework.util.AntPathMatcher.DEFAULT_PATH_SEPARATOR;
@@ -289,7 +290,8 @@ public class SwaggerUiConfigParameters extends AbstractSwaggerUiConfigProperties
 		if (supportedSubmitMethods != null)
 			org.springdoc.core.utils.SpringDocPropertiesUtils.put("supportedSubmitMethods", supportedSubmitMethods, params);
 		org.springdoc.core.utils.SpringDocPropertiesUtils.put(OAUTH2_REDIRECT_URL_PROPERTY, oauth2RedirectUrl, params);
-		org.springdoc.core.utils.SpringDocPropertiesUtils.put(URL_PROPERTY, url, params);
+		if (CollectionUtils.isEmpty(urls))
+			org.springdoc.core.utils.SpringDocPropertiesUtils.put(URL_PROPERTY, url, params);
 		put(URLS_PROPERTY, urls, params);
 		org.springdoc.core.utils.SpringDocPropertiesUtils.put("urls.primaryName", urlsPrimaryName, params);
 		org.springdoc.core.utils.SpringDocPropertiesUtils.put(TRY_IT_ENABLED_PROPERTY, tryItOutEnabled, params);


Mattias-Sehlstedt:
Note: I am a contributor, not a maintainer, so this is not something that is explicitly required for a merge

I would suggest investigating if it is possible to add a test that shows that the behavior is as expected. This to ensure that no regression for this behavior can occur. I might be something like [this](https://github.com/springdoc/springdoc-openapi/blob/main/springdoc-openapi-starter-webflux-ui/src/test/java/test/org/springdoc/ui/app1/SpringDocApp1RedirectLayoutTest.java) but with groups also defined?

springdoc-openapi-starter-common/src/main/java/org/springdoc/core/properties/SwaggerUiConfigParameters.java
```diff
@@ -39,6 +39,7 @@
 
 import org.apache.commons.lang3.StringUtils;
 
+import org.springdoc.core.utils.SpringDocPropertiesUtils;
```
Mattias-Sehlstedt, line 42: The import should not be necessary given that the full path is used still (or it might be nicer to have the import and remove the full path?)

bnasslahsen:
@souravCoder1,

Your PR is not passing - build is failing

springdoc-openapi-starter-common/src/main/java/org/springdoc/core/properties/SwaggerUiConfigParameters.java
souravCoder1: correct, we can remove

souravCoder1:
> @souravCoder1,
> 
> Your PR is not passing - build is failing

looks like test is filing

Error:  Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.258 s <<< FAILURE! -- in test.org.springdoc.ui.app40.SpringDocApp40Test
Error:  test.org.springdoc.ui.app40.SpringDocApp40Test.swagger_config_preserves_yaml_urls_for_multiple_groups -- Time elapsed: 0.007 s <<< FAILURE!
java.lang.AssertionError: No value at JSON path "url"
	at org.springframework.test.util.JsonPathExpectationsHelper.evaluateJsonPath(JsonPathExpectationsHelper.java:351)
	at org.springframework.test.util.JsonPathExpectationsHelper.assertValue(JsonPathExpectationsHelper.java:106)
	at org.springframework.test.web.servlet.result.JsonPathResultMatchers.lambda$value$0(JsonPathResultMatchers.java:88)
	at org.springframework.test.web.servlet.MockMvc$1.andExpect(MockMvc.java:214)
	at test.org.springdoc.ui.app40.SpringDocApp40Test.swagger_config_preserves_yaml_urls_for_multiple_groups(SpringDocApp40Test.java:42)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
Caused by: com.jayway.jsonpath.PathNotFoundException: No results for path: $['url']

we need to update the test file.

PR #3254, opened 2026-03-31
Upgrade Spring Boot from 2.5.11 to 2.5.13

pcalouche:
This updates the Spring Boot version to 3.5.13. It also removes the dependency management of `spring-security-oauth2-authorization-server` since Spring Boot's parent manages it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 	<parent>
 		<groupId>org.springframework.boot</groupId>
 		<artifactId>spring-boot-starter-parent</artifactId>
-		<version>3.5.11</version>
+		<version>3.5.13</version>
 	</parent>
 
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -95,12 +93,6 @@
 				<type>pom</type>
 				<scope>import</scope>
 			</dependency>
-			<!-- spring Security Oauth2 Authorization Server-->
-			<dependency>
-				<groupId>org.springframework.security</groupId>
-				<artifactId>spring-security-oauth2-authorization-server</artifactId>
```
pcalouche, line 101: Since this is managed by the Spring Boot parent, it can be omitted. This ensures that the version of Spring Auth Server used will align with the current Spring Boot version.

PR #3255, opened 2026-04-06
Direction enum:  fixed visibility scope of group order so that setGroupsOrder method can be used

sisco70:
Make Direction Enum **public**, so we can change group direction programmatically.
Direction is not visible outside the package **org.springdoc.core.properties**


Mattias-Sehlstedt:
Hi, I would suggest changing the commit message so that it somewhat describes what the PR changes and wants to achieve, so that it is possible to derive what the change is from the commit message.

E.g., "fix visibility scope of group order Direction enum so that setGroupsOrder method can be used".

PR #3266, opened 2026-04-13
Upgrade swagger-core from version 2.2.47 to 2.2.48

Mattias-Sehlstedt:
Changes are from [swagger-core reverting their change to the representation of an empty default value.](https://github.com/swagger-api/swagger-core/commit/935d6c4dd56899e7022d79381cdbfb9157bbba4a)


glandais-nickel:
I would have prefered a solution that would support both versions, as we can't be 100% sure that provided swagger-core is 2.2.48+ with dependency managers (and because we also have swagger-*-jakarta now ...).

Because if you have 2.2.47 with this commit in springdoc, it will put ##default in openapi :(

PR #3267, opened 2026-04-14
Fixes #3263 - strip null-key entries from component schema properties…

### LINKED ISSUE #3263: Null key for a Map not allowed in JSON
Springdoc 3.0.3 downgrade to 3.0.2 works well.

[swagger-core-null-key-reproducer.zip](https://github.com/user-attachments/files/26668870/swagger-core-null-key-reproducer.zip)

## Description of the problem/issue

When swagger-core resolves a bean property annotated with `@JsonUnwrapped`, the generated OpenAPI schema's `properties` map may contain a `null` key entry, which causes Jackson serialization to fail with `JsonMappingException: Null key for a Map not allowed in JSON`.

This is triggered by Spring HATEOAS `EntityModel<T>`, where `getContent()` is annotated with `@JsonUnwrapped`. When swagger-core resolves `EntityModel<SomeDto>`, it unwraps the DTO's properties into the parent schema. However, the unwrapped property `Schema` objects may have `name == null` due to `Schema.getName()` being `@JsonIgnore` and being lost during JSON-based clone operations within `ModelResolver`. These null-named schemas are then inserted into the properties map via `modelProps.put(prop.getName(), prop)`, producing a `null` key.

### Root Cause Analysis

The bug is in `ModelResolver.java` and involves three interacting mechanisms:

**1. `handleUnwrapped()` (line ~1461) extracts Schema values from the inner model's properties map:**

```java
private void handleUnwrapped(List<Schema> props, Schema innerModel,
                              String prefix, String suffix, List<String> requiredProps) {
    if (StringUtils.isBlank(suffix) && StringUtils.isBlank(prefix)) {
        if (innerModel.getProperties() != null) {
            props.addAll(innerModel.getProperties().values());  // Schema values added directly
            // ...
        }
    }
    // ...
}
```

**2. `Schema.getName()` is `@JsonIgnore`, lost during any JSON round-trip clone:**

```java
// io.swagger.v3.oas.models.media.Schema
@JsonIgnore
public String getName() {
    return this.name;
}
```

The `AnnotationsUtils.clone()` method only restores the name for the **top-level** schema, not for nested property schemas:

```java
public static Schema clone(Schema schema, boolean openapi31) {
    String cloneName = schema.getName();
    schema = Json.mapper().readValue(Json.pretty(schema), Schema.class);  // name lost for nested schemas
    schema.setName(cloneName);  // only top-level name is restored
    return schema;
}
```

**3. Null key insertion (line ~940):**

```java
for (Schema prop : props) {
    modelProps.put(prop.getName(), prop);   // null key if prop.getName() is null
}
```

When `handleUnwrapped` retrieves property schemas from a model that has been through a clone/re-resolution cycle (e.g., `$ref` dereferencing at line ~817, `resolveSubtypes`, or `context.defineModel` + later `getDefinedModels` retrieval), the nested property `Schema` objects have `name == null`. These are added to the `props` list by `handleUnwrapped`, and then `modelProps.put(prop.getName(), prop)` inserts a `null` key into the `LinkedHashMap`.

### Suggested Fix

In `handleUnwrapped()`, when no prefix/suffix is specified, preserve the property name from the inner model's properties map key instead of relying on `Schema.getName()`:

```java
private void handleUnwrapped(List<Schema> props, Schema innerModel,
                              String prefix, String suffix, List<String> requiredProps) {
    if (StringUtils.isBlank(suffix) && StringUtils.isBlank(prefix)) {
        if (innerModel.getProperties() != null) {
            for (Map.Entry<String, Schema> entry : innerModel.getProperties().entrySet()) {
                Schema prop = entry.getValue();
                if (prop.getName() == null) {
                    prop.setName(entry.getKey());  // restore name from map key
                }
                props.add(prop);
            }
            if (innerModel.getRequired() != null) {
                requiredProps.addAll(innerModel.getRequired());
            }
        }
    }
    // ...
}
```

Alternatively, the null key insertion point at line ~940 could be guarded:

```java
for (Schema prop : props) {
    if (prop.getName() != null) {
        modelProps.put(prop.getName(), prop);
    }
}
```

## Affected Version

- swagger-core-jakarta: **2.2.47**
- swagger-models-jakarta: **2.2.47**

Earliest version the bug appears in (if known): likely any version that includes the `handleUnwrapped` method using `props.addAll(innerModel.getProperties().values())` combined with `@JsonIgnore` on `Schema.getName()`.

## Steps to Reproduce

1. Add swagger-core-jakarta 2.2.47 (e.g., via springdoc-openapi-starter-webmvc-ui 3.0.3) to a Spring Boot project that also uses Spring HATEOAS.

2. Define a DTO:

```java
public record MyDto(String name, int count) {}
```

3. Define a REST endpoint returning `EntityModel<MyDto>`:

```java
@GetMapping("/items")
public PagedModel<EntityModel<MyDto>> getItems(PagedResourcesAssembler<MyDto> assembler, Pageable pageable) {
    return assembler.toModel(repository.findAll(pageable));
}
```

Spring HATEOAS `EntityModel.getContent()` is annotated with `@JsonUnwrapped`, which triggers the unwrapped schema resolution path in `ModelResolver`.

4. Access the OpenAPI docs endpoint: `GET /v3/api-docs`

5. The request fails with a `JsonMappingException` because the generated schema for `EntityModelMyDto` has a `null` key in its `properties` map.

## Expected Behavior

The OpenAPI schema for `EntityModelMyDto` should contain properly keyed properties (e.g., `name`, `count`, `_links`) with no `null` entries in the properties map.

## Actual Behavior

The `properties` map of the `EntityModelMyDto` schema contains a `null` key, causing Jackson to fail during serialization of the OpenAPI spec:

```
JsonMappingException: Null key for a Map not allowed in JSON (use a converting NullKeySerializer?)
(through reference chain:
  io.swagger.v3.oas.models.OpenAPI["components"]
  ->io.swagger.v3.oas.models.Components["schemas"]
  ->java.util.LinkedHashMap["EntityModelMyDto"]
  ->io.swagger.v3.oas.models.media.JsonSchema["properties"]
  ->java.util.LinkedHashMap["null"])
```

## Logs / Stack Traces

```
com.fasterxml.jackson.databind.JsonMappingException: Null key for a Map not allowed in JSON (use a converting NullKeySerializer?) (through reference chain: io.swagger.v3.oas.models.OpenAPI["components"]->io.swagger.v3.oas.models.Components["schemas"]->java.util.LinkedHashMap["EntityModelContractResponse"]->io.swagger.v3.oas.models.media.JsonSchema["properties"]->java.util.LinkedHashMap["null"])
        at com.fasterxml.jackson.databind.JsonMappingException.from(JsonMappingException.java:282)
        at com.fasterxml.jackson.databind.SerializerProvider.mappingException(SerializerProvider.java:1417)
        at com.fasterxml.jackson.databind.SerializerProvider.reportMappingProblem(SerializerProvider.java:1315)
        at com.fasterxml.jackson.databind.ser.impl.FailingSerializer.serialize(FailingSerializer.java:31)
        at com.fasterxml.jackson.databind.ser.std.MapSerializer.serializeOptionalFields(MapSerializer.java:867)
        at com.fasterxml.jackson.databind.ser.std.MapSerializer.serializeWithoutTypeInfo(MapSerializer.java:759)
        at com.fasterxml.jackson.databind.ser.std.MapSerializer.serialize(MapSerializer.java:719)
        at com.fasterxml.jackson.databind.ser.std.MapSerializer.serialize(MapSerializer.java:34)
        at com.fasterxml.jackson.databind.ser.BeanPropertyWriter.serializeAsField(BeanPropertyWriter.java:732)
        at com.fasterxml.jackson.databind.ser.std.BeanSerializerBase.serializeFields(BeanSerializerBase.java:760)
        at com.fasterxml.jackson.databind.ser.BeanSerializer.serialize(BeanSerializer.java:183)
        at io.swagger.v3.core.jackson.Schema31Serializer.serialize(Schema31Serializer.java:51)
        at io.swagger.v3.core.jackson.Schema31Serializer.serialize(Schema31Serializer.java:12)
        at com.fasterxml.jackson.databind.ser.std.MapSerializer.serializeOptionalFields(MapSerializer.java:868)
        at com.fasterxml.jackson.databind.ser.std.MapSerializer.serializeWithoutTypeInfo(MapSerializer.java:759)
        at com.fasterxml.jackson.databind.ser.std.MapSerializer.serialize(MapSerializer.java:719)
        at com.fasterxml.jackson.databind.ser.std.MapSerializer.serialize(MapSerializer.java:34)
```

### Environment

- Spring Boot 4.0.5
- Spring HATEOAS 3.0.3
- springdoc-openapi-starter-webmvc-ui 3.0.3
- swagger-core-jakarta 2.2.47
- Kotlin 2.3.20

### Workaround

Register a springdoc `OpenApiCustomizer` bean to strip null keys from all schema properties maps after generation:

```java
@Bean
public OpenApiCustomizer nullKeyCleanupCustomizer() {
    return openApi -> {
        if (openApi.getComponents() != null && openApi.getComponents().getSchemas() != null) {
            openApi.getComponents().getSchemas().values().forEach(this::removeNullKeys);
        }
    };
}

private void removeNullKeys(Schema<?> schema) {
    if (schema.getProperties() != null) {
        schema.getProperties().entrySet().removeIf(e -> e.getKey() == null);
        schema.getProperties().values().forEach(this::removeNullKeys);
    }
}
```

### Key code references in swagger-core 2.2.47

| File | Line(s) | Description |
|---|---|---|
| `ModelResolver.java` | ~796-812 | `jsonUnwrappedHandler` lambda registration |
| `ModelResolver.java` | ~940-941 | `modelProps.put(prop.getName(), prop)` — null key insertion point |
| `ModelResolver.java` | ~1461-1489 | `handleUnwrapped()` — extracts property schemas relying on `Schema.getName()` |
| `Schema.java` | `getName()` | `@JsonIgnore` annotation causes name loss during JSON round-trip clone |
| `AnnotationsUtils.java` | `clone()` | Only restores top-level schema name, not nested property names |

## Checklist

- [X] I have searched the [existing issues](https://github.com/swagger-api/swagger-core/issues) and this is not a duplicate.
- [x] I have provided sufficient information for maintainers to reproduce the issue.

Shaveenblu:
When swagger-core resolves a @JsonUnwrapped bean property (e.g. Spring HATEOAS EntityModel<T>), it uses JSON-based cloning internally. Schema.getName() is @JsonIgnore, so the name is lost during cloning. These null-named schemas are then inserted as null keys into the properties map, causing Jackson to throw JsonMappingException: Null key for a Map not allowed in JSON when serializing the OpenAPI document.

Add SpringDocUtils.removeNullKeySchemas(OpenAPI) which recursively removes any null-key entry from every component schema properties map, and call it in AbstractOpenApiResource.getOpenApi() after schema resolution and before the user-facing customizers run.

Five unit tests added in SpringDocUtilsTest covering top-level null keys, nested null keys, null OpenAPI, null components, and schemas with no properties.

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java
@@ -407,6 +407,8 @@ public abstract class AbstractOpenApiResource extends SpecFilter {
 				if (springDocConfigProperties.isRemoveBrokenReferenceDefinitions())
 					this.removeBrokenReferenceDefinitions(openAPI);
 
+				SpringDocUtils.removeNullKeySchemas(openAPI);
+
 				// run the optional customizers
 				List<Server> servers = openAPI.getServers();
 				List<Server> serversCopy = cloneViaJson(servers,  new TypeReference<List<Server>>() {},  springDocProviders.jsonMapper());
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/utils/SpringDocUtils.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/utils/SpringDocUtils.java
@@ -37,6 +37,7 @@ import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.swagger.v3.core.converter.AnnotatedType;
 import io.swagger.v3.core.util.PrimitiveType;
+import io.swagger.v3.oas.models.OpenAPI;
 import io.swagger.v3.oas.models.media.ComposedSchema;
 import io.swagger.v3.oas.models.media.Content;
 import io.swagger.v3.oas.models.media.Schema;
@@ -213,6 +214,36 @@ public class SpringDocUtils {
 		if (schema.getProperties() != null) {
 			schema.getProperties().values().forEach(prop -> fixNullOnlyAdditionalProperties((Schema<?>) prop));
 		}
+	}
+
+	/**
+	 * Remove null-key entries from all component schema properties maps.
+	 * Guards against a swagger-core bug where @JsonUnwrapped properties can produce
+	 * null-named schemas that get inserted as null keys, causing Jackson serialization to fail.
+	 *
+	 * @param openAPI the open api
+	 */
+	public static void removeNullKeySchemas(OpenAPI openAPI) {
+		if (openAPI == null || openAPI.getComponents() == null
+				|| openAPI.getComponents().getSchemas() == null) {
+			return;
+		}
+		openAPI.getComponents().getSchemas().values()
+				.forEach(schema -> removeNullKeyFromSchemaProperties((Schema<?>) schema));
+	}
+
+	/**
+	 * Recursively remove null-key entries from the given schema's properties map.
+	 *
+	 * @param schema the schema
+	 */
+	private static void removeNullKeyFromSchemaProperties(Schema<?> schema) {
+		if (schema == null || schema.getProperties() == null) {
+			return;
+		}
+		schema.getProperties().entrySet().removeIf(entry -> entry.getKey() == null);
+		schema.getProperties().values()
+				.forEach(prop -> removeNullKeyFromSchemaProperties((Schema<?>) prop));
 	}
 
 	/**
--- a/springdoc-openapi-starter-common/src/test/java/org/springdoc/core/utils/SpringDocUtilsTest.java
+++ b/springdoc-openapi-starter-common/src/test/java/org/springdoc/core/utils/SpringDocUtilsTest.java
@@ -1,0 +1,104 @@
+/*
+ *
+ *  *
+ *  *  *
+ *  *  *  *
+ *  *  *  *  * Copyright 2019-2026 the original author or authors.
+ *  *  *  *  *
+ *  *  *  *  * Licensed under the Apache License, Version 2.0 (the "License");
+ *  *  *  *  * you may not use this file except in compliance with the License.
+ *  *  *  *  * You may obtain a copy of the License at
+ *  *  *  *  *
+ *  *  *  *  *      https://www.apache.org/licenses/LICENSE-2.0
+ *  *  *  *  *
+ *  *  *  *  * Unless required by applicable law or agreed to in writing, software
+ *  *  *  *  * distributed under the License is distributed on an "AS IS" BASIS,
+ *  *  *  *  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  *  *  *  * See the License for the specific language governing permissions and
+ *  *  *  *  * limitations under the License.
+ *  *  *  *
+ *  *  *
+ *  *
+ *
+ */
+
+package org.springdoc.core.utils;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import io.swagger.v3.oas.models.Components;
+import io.swagger.v3.oas.models.OpenAPI;
+import io.swagger.v3.oas.models.media.Schema;
+import io.swagger.v3.oas.models.media.StringSchema;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests for {@link SpringDocUtils}.
+ */
+class SpringDocUtilsTest {
+
+	@Test
+	void removeNullKeySchemas_removesNullKeyFromTopLevelSchemaProperties() {
+		Schema<Object> parentSchema = new Schema<>();
+		Map<String, Schema> props = new LinkedHashMap<>();
+		props.put("name", new StringSchema());
+		props.put(null, new StringSchema());
+		props.put("count", new StringSchema());
+		parentSchema.setProperties(props);
+
+		OpenAPI openAPI = new OpenAPI()
+				.components(new Components().addSchemas("MyDto", parentSchema));
+
+		SpringDocUtils.removeNullKeySchemas(openAPI);
+
+		Map<String, Schema> result = openAPI.getComponents().getSchemas().get("MyDto").getProperties();
+		assertThat(result).containsOnlyKeys("name", "count");
+		assertThat(result).doesNotContainKey(null);
+	}
+
+	@Test
+	void removeNullKeySchemas_removesNullKeyFromNestedSchemaProperties() {
+		Schema<Object> nestedSchema = new Schema<>();
+		Map<String, Schema> nestedProps = new LinkedHashMap<>();
+		nestedProps.put("field", new StringSchema());
+		nestedProps.put(null, new StringSchema());
+		nestedSchema.setProperties(nestedProps);
+
+		Schema<Object> parentSchema = new Schema<>();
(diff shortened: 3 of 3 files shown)


Mattias-Sehlstedt:
Hi, have we investigated the cause for this issue? My understanding from the github-issue is that it is entirely new with 3.0.3, and thus that it is a regression stemming from somewhere. If that is the case, I would find it suitable to first look for the cause of the regression, since it could affect many other things, before we introduce a "cleaner" that simply attempts to cover the issue up with additional post-processing.

It might for example be tied to this [new way of conducting a copy](https://github.com/springdoc/springdoc-openapi/commit/05957ce95d0deeb054a3ddbcddc1dd03c6f25f5f#diff-92f9ae553c63131f1baf6c9188ab1c9bcb468ca81dfe88358256abefa23af802R251). But it is of course easiest to test verify if we use Springdoc 3.0.2 and then bump only swagger-core (to tell if the change merely comes from the bump of swagger-core or if the issue resides in springdoc itself).

Shaveenblu:
> Hi, have we investigated the cause for this issue? My understanding from the github-issue is that it is entirely new with 3.0.3, and thus that it is a regression stemming from somewhere. If that is the case, I would find it suitable to first look for the cause of the regression, since it could affect many other things, before we introduce a "cleaner" that simply attempts to cover the issue up with additional post-processing.
> 
> It might for example be tied to this [new way of conducting a copy](https://github.com/springdoc/springdoc-openapi/commit/05957ce95d0deeb054a3ddbcddc1dd03c6f25f5f#diff-92f9ae553c63131f1baf6c9188ab1c9bcb468ca81dfe88358256abefa23af802R251). But it is of course easiest to test verify if we use Springdoc 3.0.2 and then bump only swagger-core (to tell if the change merely comes from the bump of swagger-core or if the issue resides in springdoc itself).

Hey, traced the diff between swagger-core 2.2.46 and 2.2.47. The regression comes from the bump, not from springdoc. handleUnwrapped() itself didn't change, but PR #5079 changed annotation stripping for container items to also strip NULLABLE_ANNOTATIONS. EntityModel.getContent() is annotated with both @Nullable and @JsonUnwrapped, so that change altered its inner model resolution path pushing it through a JSON clone cycle where Schema.getName() (which is @JsonIgnore) gets lost. Those null-named schemas then hit modelProps.put(prop.getName(), prop) producing the null key. Nothing changed in springdoc between 3.0.2 and 3.0.3 other than the version bump. Upstream issue is swagger-api/swagger-core#5126. PR #3266 (2.2.48) doesn't fix it. Happy to close this and wait for the upstream fix or keep it as a short-term workaround up to you.

Mattias-Sehlstedt:
Thanks for the additional information.

I am the author of #5079,. so if that change is the cause of this, then I believe that the best approach is for me to revisit that change and implement it differently so that it does not cause there changes (and with that also document this behavior with tests so that we know how it is used). I can investigate further on my end, but can make no promises with regard to when I have time.

You can leave this PR up since I am only a contributor. The maintainer can decide what they consider the best and easiest approach.

Mattias-Sehlstedt:
Since you have worked on providing a solution for the reported issue, could you aid me in understanding how to properly reproduce the issue? See https://github.com/springdoc/springdoc-openapi/issues/3263#issuecomment-4315498801.

bnasslahsen:
closed, for conflicts and no activity on this one for several months

PR #3269, opened 2026-04-20
Add mechanism to disable nullable for Kotlin properties

0xabadea:
The newly-added `KotlinNullablePropertyCustomizer` results in Kotlin nullable properties being always marked as non-required *and* nullable. Sometimes that may not be a desirable output.

The Java/Kotlin+Jackson ecosystem does not generally or easily support undefined *and* null properties at the same time. Usually, Kotlin properties with a null value are either serialized as `"property": null` or they are left out (making them undefined). In our backend we leave them out via `@JsonInclude` `non_null/non_absent` and we document them as non-required in our OpenAPI spec. The understanding is that they may be missing from the payload, but if they are present, they are never null.

Our frontend then generates TypeScript types with optional properties, as in `property?: TheType`. In this setup, marking these properties as nullable in the OpenAPI spec results in optional *and* nullable TypeScript properties, as in `property?: TheType | null`. This brings no additional value and is actually detrimental, as the frontend developers now need to handle null values that never occur in practice.

I propose the introduction of a new property `springdoc.model-converters.kotlin-nullable-property-customizer.enabled` (defaulted to true) that would allow disabling `KotlinNullablePropertyCustomizer` on demand.

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/configuration/SpringDocKotlinConfiguration.kt
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/configuration/SpringDocKotlinConfiguration.kt
@@ -81,6 +81,7 @@ class SpringDocKotlinConfiguration() {
 		@Bean
 		@Lazy(false)
 		@ConditionalOnMissingBean
+		@ConditionalOnProperty(name = [Constants.SPRINGDOC_KOTLIN_NULLABLE_PROPERTY_CUSTOMIZER_ENABLED], matchIfMissing = true)
 		fun kotlinNullablePropertyCustomizer(objectMapperProvider: ObjectMapperProvider): KotlinNullablePropertyCustomizer {
 			return KotlinNullablePropertyCustomizer(objectMapperProvider)
 		}
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/utils/Constants.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/utils/Constants.java
@@ -107,6 +107,11 @@ public final class Constants {
 	public static final String SPRINGDOC_POLYMORPHIC_CONVERTER_ENABLED = "springdoc.model-converters.polymorphic-converter.enabled";
 
 	/**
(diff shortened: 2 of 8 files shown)


fatso83:
This is highly anticipated 😉

thejeff77:
For visibility on the wider plan: rather than landing a global flag for the customizer, I've proposed reverting #3256 (#3276) and adding a proper per-property override upstream in swagger-core.

- Issue: swagger-api/swagger-core#5160
- Implementation PR: swagger-api/swagger-core#5161 — `Schema.NullableMode { AUTO, NULLABLE, NOT_NULLABLE }` mirroring the established `RequiredMode` / `AccessMode` pattern.

Rationale (full discussion on #3256): a global flag addresses the symptom, but the deeper bug is that users have no way to opt a single property out of auto-detection — and `@Schema(nullable = false)` cannot be honored because the boolean defaults to `false` and is indistinguishable from "not specified" via reflection. `NullableMode` fixes that the same way `RequiredMode` did for `required` in 2022. Acknowledging your work and pushback here directly led to this conclusion.

0xabadea:
Hello @bnasslahsen, this PR has not been getting any attention so far.

Is there anything I can do to get merged either this or https://github.com/springdoc/springdoc-openapi/pull/3276 before the next version is released?

bnasslahsen:
@0xabadea, 
Thats'ok - fix merged

rupert-jung-mw:
So, kotlin-nullable-property-customizer is not existing/working in the current version 3.1.0 but is actually listed as added feature...? I am still stuck to 3.0.2 and the proposal for swagger-core has been backlogged...

Mattias-Sehlstedt:
@rupert-jung-mw I would suggest to raise a separate issue and also share a reproducible scenario for it. From what I can see the tests introduced in this PR correctly illustrates that the setting has an effect.

0xabadea:
Hi @rupert-jung-mw, if you remember to ping me in the new issue I will take a look at it.

rupert-jung-mw:
Follow up: Updated to 3.1.1 and added property:

`springdoc.model-converters.kotlin-nullable-property-customizer.enabled: false`

I can confirm that `| null` is not added to non-required properties anymore :)

Issue #3274, opened 2026-04-21
Description disappears from the generated json after upgrade to 3.0.3


petergphillips:
**Describe the bug**
After the upgrade from springdoc 3.0.2 to 3.0.3 we've noticed that
```kotlin
  @field:Schema(description = "The object of the person to greet")
  val objectName: SomeObject? = null,
```
gets generated as
```json
          "objectName" : {
            "oneOf" : [ {
              "$ref" : "#/components/schemas/SomeObject"
            }, {
              "type" : "null"
            } ]
          }
```
and the `description` is then not set at all.

Under springdoc 3.0.2 the same code generated:
```json
          "objectName" : {
            "$ref" : "#/components/schemas/SomeObject",
            "description" : "The object of the person to greet"
          }
```

**To Reproduce**
- What version of spring-boot you are using? 4.0.5
- What version of kotlin are you using? 2.3.20
- What modules and versions of springdoc-openapi are you using? 3.0.3
- What is the actual and the expected result using OpenAPI Description (yml or json)? See above for actual and also the generated code in 3.0.2. I would expect the `description` to be still set.
- Provide with a sample code (HelloController) or Test that reproduces the problem
```kotlin
@RestController
class IssuesController {
  @GetMapping("/")
  fun greet(greeting: Greeting): String = greeting.objectName?.toString() ?: "Hello World"
}

data class Greeting(
  @field:Schema(description = "The object of the person to greet")
  val objectName: SomeObject? = null,
)
data class SomeObject(val someProperty: String)
```

**Expected behavior**
The `description` is generated and not missing.

Note that if the field is defined as
```kotlin
  val stringName: String? = null,
```
then the generated code is
```json
          "stringName" : {
            "type" : [ "string", "null" ],
            "description" : "The name of the person to greet"
          },
```

petergphillips:
See https://github.com/petergphillips/springdoc-issues/blob/pgp-springdoc-3274/README.md for a sample project and test demonstrating the issue.

petergphillips:
Also just discovered that `deprecated` disappears as well. So:
```kotlin
  @Deprecated("Use stringName instead")
  val objectName: SomeObject? = null,
```
used to generate
```json
          "objectName" : {
            "$ref" : "#/components/schemas/SomeObject",
            "deprecated" : true,
          }
```
and now just generates
```json
          "objectName" : {
            "oneOf" : [ {
              "$ref" : "#/components/schemas/SomeObject"
            }, {
[...]

PR #3276, opened 2026-04-23
revert: auto-set nullable for Kotlin nullable types (#3256)

thejeff77:
## Summary

Reverts #3256. The `KotlinNullablePropertyCustomizer` auto-marks Kotlin `T?` as nullable with no per-property opt-out, which silently produces incorrect specs for backends using `@JsonInclude(NON_NULL)`. This restores pre-#3256 behavior pending an upstream `NullableMode` proposal in swagger-core.

## Why revert rather than patch

The customizer auto-marks any Kotlin `T?` property as nullable in the generated schema. This is correct *only* when Jackson actually serializes null values onto the wire — i.e., default Jackson configuration. For backends that use `@JsonInclude(NON_NULL)` (a widespread Jackson convention) the resulting spec is wrong: it claims null can appear when in fact Jackson strips null fields entirely.

The natural per-property override would be `@Schema(nullable = false)`. This cannot be honored. `@Schema.nullable()` is a `boolean` defaulting to `false`, which means via reflection there is no way to distinguish "user explicitly wrote `nullable = false`" from "user never specified anything." This is a known design wart on `@Schema`'s legacy boolean fields — the same issue swagger-core addressed for `required` by introducing `RequiredMode` in 2022 (swagger-api/swagger-core#4221, swagger-api/swagger-core#4286, commit `b1729fcf1`) and for `readOnly`/`writeOnly` by introducing `AccessMode` in 2018 (swagger-api/swagger-core#2675, swagger-api/swagger-core#2677). No equivalent `NullableMode` exists yet.

Without that override, this PR's auto-detection cannot be made correct for the population it impacts. A global enable/disable flag (proposed in #3269) addresses the symptom but not the underlying override gap — codebases that mix `@JsonInclude(NON_NULL)` and default-Jackson DTOs still have no clean answer, and shipping the flag adds global config surface for a problem that will be solved per-property upstream once `NullableMode` exists.

## Path forward

1. **This revert** — restores pre-#3256 behavior. Users in the original #906 cohort fall back to the explicit `@field:Schema(nullable = true)` annotation, the long-standing workaround they were using before #3256.
2. **Swagger-core proposal for `NullableMode { AUTO, NULLABLE, NOT_NULLABLE }`** mirroring `RequiredMode`. Issue/PR to be opened on swagger-api/swagger-core. This benefits both layers — the swagger-core `@Nullable` auto-detection added in swagger-api/swagger-core#5018 (2025-11) hits the same override gap and would also use the new mode.
3. **Re-introduce `KotlinNullablePropertyCustomizer` once `NullableMode` ships upstream**, with the customizer honoring `NullableMode.NOT_NULLABLE` as an explicit per-property override and `NullableMode.AUTO` (default) as the auto-detection trigger.

## Background discussions

- Original feature request: #906 (first raised 2020, closed)
- User reports of breakage: comments on #3256 from @Majstr, @0xabadea, @fatso83, @rupert-jung-mw
- Proposed flag-based workaround: #3269 (close in favor of this revert)
- Related swagger-core nullable work: swagger-api/swagger-core#5018 (native `@Nullable` support), swagger-api/swagger-core#5077, swagger-api/swagger-core#5158, swagger-api/swagger-core#5124

## Test plan

- [ ] CI passes (the revert deletes the customizer and its tests; no new tests required)
- [ ] Manual confirm: a Kotlin data class with `val foo: String? = null` no longer auto-marks as `nullable: true` (back to pre-#3256 behavior)
- [ ] Manual confirm: existing `@field:Schema(nullable = true)` opt-in still works

## Risk

Low. This is a literal `git revert` of the merge commit; no manual conflict resolution. Users relying on the auto-detection lose it and must re-add `@field:Schema(nullable = true)` until the upstream `NullableMode` work lands.


thejeff77:
Cross-link: the upstream `NullableMode` proposal referenced in the path-forward section is now open at swagger-api/swagger-core#5160.

thejeff77:
Note on CI: the failing \`Java CI with Maven\` check (\`Non-resolvable parent POM for org.springdoc:springdoc-openapi-tests\`) is **pre-existing infrastructure**, not caused by this revert. Same exact failure on:

- the previous merge to \`main\` (run [24686358254](https://github.com/springdoc/springdoc-openapi/actions/runs/24686358254), 2026-04-20)
- run [24317160951](https://github.com/springdoc/springdoc-openapi/actions/runs/24317160951) (2026-04-12, on the 3.0.3 line)
- multiple intermediate \`action_required\` runs

This PR is a literal \`git revert -m 1\` of the #3256 merge commit — no manual conflict resolution, no test changes beyond the original PR's tests being deleted alongside the customizer they covered. The same code (pre-#3256) was passing CI before #3256 was introduced.

thejeff77:
@bnasslahsen — flagging this revert in case it slipped your queue. It reverts #3256 (which you merged in April) due to a missing per-property override mechanism producing silently incorrect specs for users with \`@JsonInclude(NON_NULL)\` configurations (see comments on #3256 from @0xabadea, @Majstr, @fatso83, @rupert-jung-mw). The proper upstream fix (\`Schema.NullableMode\`) is implemented at swagger-api/swagger-core#5161 and will let us re-introduce the customizer correctly once it lands. The build CI failure on this PR is the pre-existing Maven infra issue (documented above), not caused by the revert. No rush, just flagging since users on #3256 are blocked.

bnasslahsen:
@thejeff77,

should be fixed with https://github.com/springdoc/springdoc-openapi/issues/3274

PR #3280, opened 2026-04-26
Align test module parent versions with spring-boot-4 snapshot

limehee:
## Summary

Aligns the `springdoc-openapi-tests` reactor parent versions with the current `spring-boot-4` root snapshot version (`3.0.4-SNAPSHOT`).

## Background

The `spring-boot-4` root project now declares `3.0.4-SNAPSHOT`, but the test reactor modules still inherited `3.0.3-SNAPSHOT`. Since the Maven CI build activates the `ci` profile and includes `springdoc-openapi-tests`, Maven fails while reading the reactor before any tests can run.

This keeps the test modules on the same snapshot as the branch root so the local parent POM resolves correctly.

## Verification

- `mvn -B validate --file pom.xml`

### DIFF
--- a/springdoc-openapi-tests/pom.xml
+++ b/springdoc-openapi-tests/pom.xml
@@ -2,7 +2,7 @@
 	<parent>
 		<artifactId>springdoc-openapi</artifactId>
 		<groupId>org.springdoc</groupId>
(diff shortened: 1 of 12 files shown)


bnasslahsen:
@limehee,

Thanks for your proposal. Alignement already applied!

PR #3312, opened 2026-08-03
Upgrade swagger-core from version 2.2.52 to 2.2.53

Mattias-Sehlstedt:
Upgrades swagger-core from 2.2.52 to 2.2.53, and with that upgrade some local bug-fixes are removed since they have been fixed upstream.

Changes:
removal of local method `fixOAS31ExclusiveConstraints` is tied to https://github.com/swagger-api/swagger-core/issues/5170
removal of local method `fixNullOnlyAdditionalProperties` is tied to https://github.com/swagger-api/swagger-core/issues/5115
https://github.com/springdoc/springdoc-openapi/commit/9ed23400200f22213477fc8f6a4494578369d56c is removed due to introduction of https://github.com/swagger-api/swagger-core/pull/5192
https://github.com/springdoc/springdoc-openapi/commit/4a5cbf52ba865e03bb5379c5dd70b66dc98e1767 is removed due to introduction of https://github.com/swagger-api/swagger-core/pull/5193


bender316:
@Mattias-Sehlstedt Could you update your PR to use swagger-core:2.2.54 in order to get the fix for swagger-api/swagger-core#5226 into springdoc.

Mattias-Sehlstedt:
I would prefer to keep this PR strongly targeted against the upstream behavior that replaces local fixes. This so that it will be clear where issues come from if there are any. But I can create another PR for an additional bump once this one is merged (might be that swagger-core does another release before this one is handled too).

PR #3315, opened 2026-08-06
fix: Add a type deserializer to allow for JSON cloning of a JsonSchema object

### LINKED ISSUE #3314: SpringDocUtils logs Json Processing Exception occurred warning for every @Min-constrained Integer query parameter (regression since 3.1.0)
**Describe the bug**

  Starting with 3.1.0, any endpoint with an `Integer`/`int` query parameter carrying a Bean Validation constraint (e.g. `@Min`) causes a WARN log on every api-docs request:

  Json Processing Exception occurred: Cannot construct instance of java.util.HashSet (although at least one Creator exists): no String-argument constructor/factory method to deserialize from String value ('integer')
   at [Source: ...]; line: 1, column: 9] (through reference chain: io.swagger.v3.oas.models.media.JsonSchema["type"])

  It also occurs with `array`-typed schemas (`ArraySchema["type"]`, value `'array'`). On a real API with many endpoints this produces dozens of these warnings per api-docs request.

  **To Reproduce**

  - Spring Boot 4.1.0
  - `springdoc-openapi-starter-webmvc-ui` 3.1.0

  I wrote a minimal Spring Boot app and a test that calls `GET /v3/api-docs` and asserts no `SpringDocUtils` WARN is logged. I ran it myself against both versions to confirm the regression:

  ```java
  @RestController
  class HelloController {
      @GetMapping("/hello")
      String hello(@RequestParam(required = false) @Min(1) Integer limit) {
          return "hello";
      }
  }

  @SpringBootTest
  @AutoConfigureMockMvc
  class ApiDocsWarningTest {

      @Autowired
      MockMvc mockMvc;

      ListAppender<ILoggingEvent> appender;

      @BeforeEach
      void attachAppender() {
          appender = new ListAppender<>();
          appender.start();
          ((Logger) LoggerFactory.getLogger("org.springdoc.core.utils.SpringDocUtils"))
                  .addAppender(appender);
      }

      @Test
      void apiDocsRequestLogsNoJsonProcessingWarning() throws Exception {
          mockMvc.perform(get("/v3/api-docs"))
                  .andExpect(status().isOk());

          List<String> warnMessages = appender.list.stream()
                  .filter(e -> e.getLevel() == Level.WARN)
                  .map(ILoggingEvent::getFormattedMessage)
                  .toList();

          assertThat(warnMessages)
                  .noneMatch(m -> m.contains("Json Processing Exception occurred"));
      }
  }
```
  
  - With springdoc-openapi-starter-webmvc-ui:3.0.3 → test passes, no warning.
  - With springdoc-openapi-starter-webmvc-ui:3.1.0, everything else unchanged → test fails, SpringDocUtils logs the Json Processing Exception occurred warning shown above.

  **Expected behavior**

  No warning should be logged when generating the OpenAPI docs for this endpoint.

  **Additional context**

  - Response is still 200 OK with an apparently valid spec, so functionally nothing is visibly broken — it's just log noise.
  - Bisected against every version between 3.0.1 and 3.1.0 published on Maven Central (no releases exist between 3.0.3 and 3.1.0, it's a direct jump). 3.0.1, 3.0.2, 3.0.3 are all clean; 3.1.0 is the first version to show this.

Mattias-Sehlstedt:
Change https://github.com/springdoc/springdoc-openapi/commit/47002081dde26868d36b43a2ec1577ac408cd01b introduced logic that involved cloning a schema with `schema = cloneViaJson(schema, schema.getClass(), parameterBuilder.getObjectMapperProvider().jsonMapper());`. This logic will fail for an OAS 3.1 specification, since the `JsonSchema` from swagger-core is not symmetric with regard to the `type(s)` field when doing back and forth de/serialization. See [serialization configuration](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-core/src/main/java/io/swagger/v3/core/jackson/mixin/Schema31Mixin.java#L65).

This means that `types: [integer]` becomes `type: integer`, which then cannot be deserialized back into `JsonSchema`.

This PR adjusts for that issue by adding a `JsonSchema` `Mixin` that forwards the `type` field into `types`.

Fixes https://github.com/springdoc/springdoc-openapi/issues/3314

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/deserializers/TypeSetDeserializer.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/deserializers/TypeSetDeserializer.java
@@ -1,0 +1,41 @@
+package org.springdoc.core.deserializers;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.JsonNode;
+
+import java.io.IOException;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+/**
+ * Reads an OpenAPI 3.1 schema "type" that swagger-core serializes as a scalar
+ * {@code "type": "integer"} back into the Set<String> field, while still supporting
+ * the array form {@code "type": ["string","null"]}.
+ * <p>
+ * Reported upstream with <a href="https://github.com/swagger-api/swagger-core/issues/5264">swagger-core issue 5264</a>
+ */
+public class TypeSetDeserializer extends JsonDeserializer<Set<String>> {
+
+    @Override
+    public Set<String> deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
+        JsonNode node = ctxt.readTree(p);
+        if (node == null || node.isNull()) {
+            return null;
+        }
+        Set<String> types = new LinkedHashSet<>();
+        if (node.isArray()) {
+            node.forEach(n -> {
+                if (!n.isNull()) {
+                    types.add(n.asText());
+                }
+            });
+        }
+        else {
+            types.add(node.asText());
+        }
+        return types.isEmpty() ? null : types;
+    }
+
+}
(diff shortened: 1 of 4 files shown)


Mattias-Sehlstedt:
Could also potentially be handled by changing
`schema = cloneViaJson(schema, schema.getClass(), parameterBuilder.getObjectMapperProvider().jsonMapper());`
 to 
 `schema = cloneViaJson(schema, Schema.class, parameterBuilder.getObjectMapperProvider().jsonMapper());`
given that we trust the type casting to not lose any necessary information. That is what is done for `cloneViaJson` for list items, and also seem to be the general pattern used within `swagger-core` as well.

PR #3323, opened 2026-08-12
feat: Add login endpoint configuration with example values for username and password

### LINKED ISSUE #3321: Support adding example values to the Spring Security login endpoint.
I want the login endpoint to automatically apply the properties from `spring.security.user` (the built-in properties of Spring Security), so that the login endpoint can have default values, and I don't have to manually enter the username and password every time.

```yaml
spring:
  security:
    user:
      name: admin
      password: password
```

However, it is recommended to add a property to control whether to automatically apply this, because in production environments the API documentation might be exposed to third parties, and you may want to have a specific switch to avoid exposing the default username and password. This property is recommended to have a default value of `false`, requiring manual enabling.

Mattias-Sehlstedt:
Adds two new configuration properties that let users define **example values** for the `username` and `password` fields of the auto-generated Spring Security login endpoint, so those examples show up in the OpenAPI document and Swagger UI "Try it out" form.

### New configuration properties

| Property | Description |
| --- | --- |
| `springdoc.login-endpoint.username-example` | Example value applied to the `username` field of the login request body. |
| `springdoc.login-endpoint.password-example` | Example value applied to the `password` field of the login request body. |

Example:

```properties
springdoc.show-login-endpoint=true
springdoc.login-endpoint.username-example=demouser
springdoc.login-endpoint.password-example=secret
```

The new `app13` tests are copied from the existing `app6` [tests](https://github.com/springdoc/springdoc-openapi/blob/main/springdoc-openapi-tests/springdoc-openapi-security-tests/src/test/java/test/org/springdoc/api/v30/app6/SpringDocApp6Test.java) to showcase that the previous behavior with `springdoc.show-login-endpoint=true` is not affected.

Fixes: https://github.com/springdoc/springdoc-openapi/issues/3321

An alternative design would be a single boolean flag (e.g. springdoc.login-endpoint.show-credentials=true) that reads the application's actual configured username/password (such as spring.security.user.name / spring.security.user.password) and injects them as examples. That was deliberately not chosen, as a separation of concerns as well as a safety measure. 

Documentation examples are a documentation concern; they should not be coupled to, or reach into, security configuration internals. Such a solution could easily result in the user setting the property to true in the root `application.properties`, and it thus affecting the `prod` settings too. This solution forces the user to specify the exact credentials, which could only do harm if the user has the same credentials in `test` **and** `prod`.

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/configuration/SpringDocSecurityConfiguration.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/configuration/SpringDocSecurityConfiguration.java
@@ -45,6 +45,8 @@ import org.slf4j.LoggerFactory;
 import org.springdoc.core.configuration.hints.SpringDocSecurityHints;
 import org.springdoc.core.customizers.GlobalOpenApiCustomizer;
 import org.springdoc.core.customizers.OpenApiCustomizer;
+import org.springdoc.core.properties.SpringDocConfigProperties;
+import org.springdoc.core.properties.SpringDocConfigProperties.LoginEndpoint;
 
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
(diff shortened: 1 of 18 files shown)


bnasslahsen:
Merged, thanks @Mattias-Sehlstedt.

Two small follow-ups on top of your commit:

- realigned the `buildRequestBody` continuation line to the 4-tab continuation indent Spring JavaFormat uses (it was aligned visually on the opening paren);
- recorded the feature in `CHANGELOG.md` under 3.1.1.

The same feature is being backported to the `spring-boot-3` line for 2.9.1.

PR #3330, opened 2026-08-16
Fix customizer ordering

### LINKED ISSUE #3320: Please make the customizers support `@Order`.
Currently, the internal `org.springdoc.core.customizers.SpringDocCustomizers` is as follows:

```java
public class SpringDocCustomizers implements ApplicationContextAware, InitializingBean {

	/**
	 * The Open api customisers.
	 */
	private final Optional<Set<OpenApiCustomizer>> openApiCustomizers;

	/**
	 * The Operation customizers.
	 */
	private final Optional<Set<OperationCustomizer>> operationCustomizers;
	
	// ... other code ...
	
}
```

And this is where it is used:

```java
public abstract class AbstractOpenApiResource extends SpecFilter {

	// ... other code ...
	
	protected OpenAPI getOpenApi(String serverBaseUrl, Locale locale) {
		this.reentrantLock.lock();
		try {
			final OpenAPI openAPI;
			final Locale finalLocale = selectLocale(locale);
			if (openAPIService.getCachedOpenAPI(finalLocale) == null || springDocConfigProperties.isCacheDisabled()) {

				// ... code ...

                // here
				openAPIService.getContext().getBeansOfType(OpenApiLocaleCustomizer.class).values().forEach(openApiLocaleCustomizer -> openApiLocaleCustomizer.customise(openAPI, finalLocale));
				springDocCustomizers.getOpenApiCustomizers().ifPresent(apiCustomizers -> apiCustomizers.forEach(openApiCustomizer -> openApiCustomizer.customise(openAPI)));

				// ... code ...
			}
			else {
				LOGGER.debug("Fetching openApi document from cache");
				openAPI = openAPIService.getCachedOpenAPI(finalLocale);
				openAPIService.updateServers(serverBaseUrl, openAPI);
			}
			return openAPI;
		}
		finally {
			SpringDocAnnotationsUtils.clearCache(operationParser.getJavadocProvider());
			this.reentrantLock.unlock();
		}
	}

	// ... other code ...

}
```

This approach does not support the `@Order` annotation.

I would like to use `ObjectProvider<T>` instead of `Optional<Set<T>>`.

```java
import org.springframework.beans.factory.ObjectProvider;

public class SpringDocCustomizers implements ApplicationContextAware, InitializingBean {

    /**
     * The Open api customisers.
     */
    private final ObjectProvider<OpenApiCustomizer> openApiCustomizers;

    /**
     * The Operation customizers.
     */
    private final ObjectProvider<OperationCustomizer> operationCustomizers;

    // ... other code ...

}
```

```java
public abstract class AbstractOpenApiResource extends SpecFilter {

	// ... other code ...
	
	protected OpenAPI getOpenApi(String serverBaseUrl, Locale locale) {
		this.reentrantLock.lock();
		try {
			final OpenAPI openAPI;
			final Locale finalLocale = selectLocale(locale);
			if (openAPIService.getCachedOpenAPI(finalLocale) == null || springDocConfigProperties.isCacheDisabled()) {

				// ... code ...

                // here
                openAPIService.getContext().getBeanProvider(OpenApiLocaleCustomizer.class).orderedStream().forEach(customizer -> customizer.customise(openAPI, finalLocale));
                springDocCustomizers.getOpenApiCustomizers().orderedStream().forEach(customizer -> customizer.customise(openAPI));

				// ... code ...
			}
			else {
				LOGGER.debug("Fetching openApi document from cache");
				openAPI = openAPIService.getCachedOpenAPI(finalLocale);
				openAPIService.updateServers(serverBaseUrl, openAPI);
			}
			return openAPI;
		}
		finally {
			SpringDocAnnotationsUtils.clearCache(operationParser.getJavadocProvider());
			this.reentrantLock.unlock();
		}
	}

	// ... other code ...

}
```

This way, Spring's built-in `.orderedStream()` can be used to handle the ordering of `@Order`.

Otherwise, the execution order cannot be guaranteed, because I want to execute the following code after `SpringDocSecurityConfiguration.SpringSecurityLoginEndpointConfiguration#springSecurityLoginEndpointCustomizer`:

Add example values to the default login endpoint.

```java
@Bean
// Set the lowest priority, execute last, to ensure it runs after SpringDocSecurityConfiguration.SpringSecurityLoginEndpointConfiguration#springSecurityLoginEndpointCustomizer
// Otherwise the login endpoint cannot be found
@Order
OpenApiCustomizer springSecurityLoginEndpointSchemaCustomizer(ApplicationContext applicationContext) {
    FilterChainProxy filterChainProxy = applicationContext.getBean(AbstractSecurityWebApplicationInitializer.DEFAULT_FILTER_NAME, FilterChainProxy.class);
    return openAPI -> {
        for (SecurityFilterChain filterChain : filterChainProxy.getFilterChains()) {
            Optional<UsernamePasswordAuthenticationFilter> optionalFilter =
                    filterChain.getFilters().stream()
                            .filter(UsernamePasswordAuthenticationFilter.class::isInstance)
                            .map(UsernamePasswordAuthenticationFilter.class::cast)
                            .findAny();
            if (optionalFilter.isPresent()) {
                UsernamePasswordAuthenticationFilter usernamePasswordAuthenticationFilter = optionalFilter.get();

                try {
                    RequestMatcher requestMatcher = (RequestMatcher) FieldUtils.readField(
                            usernamePasswordAuthenticationFilter,
                            "requiresAuthenticationRequestMatcher",
                            true
                    );

                    if (requestMatcher instanceof PathPatternRequestMatcher pathPatternRequestMatcher) {
                        String loginPath = getPath(pathPatternRequestMatcher);

                        Map.Entry<String, PathItem> loginPathEntry = openAPI.getPaths().entrySet().stream().filter(entry -> entry.getKey().equals(loginPath)).findAny().orElse(null);

                        if (loginPathEntry == null) {
                            return;
                        }

                        PathItem value = loginPathEntry.getValue();

                        value.readOperations().forEach(operation -> {
                            // Add example values to the login operation
                        });
                    }
                }
                catch (IllegalAccessException | ClassCastException e) {
                    // Exception escaped
                    log.trace(e.getMessage());
                }
            }
        }
    };
}
```

king-407:
## Summary

Fixes #3320.

This updates Springdoc customizer execution so customizers are applied according to Spring ordering semantics such as `@Order`.

Previously, several customizer collections were stored as `Set`s and iterated directly, so the execution order could depend on set iteration order instead of Spring's ordering rules.

## Changes

- Apply `OpenApiCustomizer` instances in Spring order.
- Apply `OperationCustomizer` instances in Spring order.
- Apply `RouterOperationCustomizer` instances in Spring order.
- Apply `DataRestRouterOperationCustomizer` instances in Spring order.
- Apply `OpenApiLocaleCustomizer` beans using Spring's ordered bean stream.
- Add a focused regression test for ordered customizer execution.

## Testing

```bash
mvn -q -pl springdoc-openapi-starter-common -Dtest=SpringDocCustomizersTest test
git diff --check

### DIFF
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/api/AbstractMultipleOpenApiResource.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/api/AbstractMultipleOpenApiResource.java
@@ -128,7 +128,7 @@ public abstract class AbstractMultipleOpenApiResource<R extends AbstractOpenApiR
 		this.groupedOpenApis.forEach(groupedOpenApi -> {
 					groupedOpenApi.addAllOpenApiCustomizer(springDocCustomizers.getGlobalOpenApiCustomizersStream().toList());
 					groupedOpenApi.addAllOperationCustomizer(springDocCustomizers.getGlobalOperationCustomizersStream().toList());
-					groupedOpenApi.addAllOpenApiMethodFilter(springDocCustomizers.getGlobalOpenApiMethodFiltersStream().toList());
+					springDocCustomizers.getGlobalOpenApiMethodFilters().ifPresent(groupedOpenApi::addAllOpenApiMethodFilter);
 				}
 		);
 
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/api/AbstractMultipleOpenApiResource.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/api/AbstractMultipleOpenApiResource.java
@@ -126,9 +126,9 @@ public abstract class AbstractMultipleOpenApiResource<R extends AbstractOpenApiR
 	@Override
 	public void afterPropertiesSet() {
 		this.groupedOpenApis.forEach(groupedOpenApi -> {
-					springDocCustomizers.getGlobalOpenApiCustomizers().ifPresent(groupedOpenApi::addAllOpenApiCustomizer);
-					springDocCustomizers.getGlobalOperationCustomizers().ifPresent(groupedOpenApi::addAllOperationCustomizer);
-					springDocCustomizers.getGlobalOpenApiMethodFilters().ifPresent(groupedOpenApi::addAllOpenApiMethodFilter);
+					groupedOpenApi.addAllOpenApiCustomizer(springDocCustomizers.getGlobalOpenApiCustomizersStream().toList());
+					groupedOpenApi.addAllOperationCustomizer(springDocCustomizers.getGlobalOperationCustomizersStream().toList());
+					groupedOpenApi.addAllOpenApiMethodFilter(springDocCustomizers.getGlobalOpenApiMethodFiltersStream().toList());
 				}
 		);
 
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/api/AbstractOpenApiResource.java
@@ -428,8 +428,8 @@ public abstract class AbstractOpenApiResource extends SpecFilter {
 				List<Server> servers = openAPI.getServers();
 				List<Server> serversCopy = cloneViaJson(servers,  new TypeReference<List<Server>>() {},  springDocProviders.jsonMapper());
 
-				openAPIService.getContext().getBeansOfType(OpenApiLocaleCustomizer.class).values().forEach(openApiLocaleCustomizer -> openApiLocaleCustomizer.customise(openAPI, finalLocale));
-				springDocCustomizers.getOpenApiCustomizers().ifPresent(apiCustomizers -> apiCustomizers.forEach(openApiCustomizer -> openApiCustomizer.customise(openAPI)));
+				openAPIService.getContext().getBeanProvider(OpenApiLocaleCustomizer.class).orderedStream().forEach(openApiLocaleCustomizer -> openApiLocaleCustomizer.customise(openAPI, finalLocale));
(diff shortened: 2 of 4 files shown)


zhujianxintian:
@king-407 Thank you for your help! Much appreciated 🙏

However, there is one thing I don't quite understand: why was the existing custom item collection API retained? I noticed that most of the places where it was used have also been modified, and it seems that using `ObjectProvider` wouldn't make much of a difference.

(I prefer using `ObjectProvider` because when reading the source code, if the type is `ObjectProvider`, it indicates that `@Order` is likely supported—this is more intuitive than `List` or `Set`. With `List` or `Set`, you only know `@Order` works if you check the call site, or if you don't read the source code and just use `@Order` directly.)

I'm curious—is this due to compatibility concerns, or is there another reason?

king-407:
@zhujianxintian Thanks for taking a look.

Yes, the main reason was to keep the change narrow and avoid changing the existing `SpringDocCustomizers` construction API in this PR.

`SpringDocCustomizers` is also created from `GroupedOpenApi` customizer collections, not only from Spring bean injection, so I kept the existing collection-based API and centralized the ordering at the point where the customizers are consumed.

That said, I agree that `ObjectProvider` communicates Spring ordering support more clearly when reading the code. If the maintainers prefer that direction, I’m happy to refactor the bean-injected customizers to use `ObjectProvider` while keeping the grouped/custom collection path working as well.

zhujianxintian:
Thank you for your reply, I understand the reason for doing this now. 

Both `ObjectProvider` and your approach can satisfy Spring's ordering support; the former is more concise but more aggressive, while the latter is more stable and compatible. 

I raised this question only because I wanted to improve readability further, but perhaps the latter is indeed more prudent at the moment. Both ways are good. 

PR #3351, opened 2026-09-06
upgrade swagger-core from 2.2.53 to 2.2.54

Mattias-Sehlstedt:
Swagger-core has implemented better handling for dates in https://github.com/swagger-api/swagger-core/pull/5184. This means that some parts of `ExtraSchemas` from https://github.com/springdoc/springdoc-openapi/commit/ec43ffbd6e316aa8fb359552f41fe2913a629fad can be removed, since they are now handled upstream.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,7 @@
 		<central-publishing-maven-plugin.version>0.11.0
 		</central-publishing-maven-plugin.version>
 		<flatten-maven-plugin.version>1.5.0</flatten-maven-plugin.version>
-		<swagger-api.version>2.2.53</swagger-api.version>
+		<swagger-api.version>2.2.54</swagger-api.version>
 		<swagger-ui.version>5.32.14</swagger-ui.version>
 		<gmavenplus-plugin.version>1.13.1</gmavenplus-plugin.version>
 		<jjwt.version>0.9.1</jjwt.version>
--- a/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/utils/SpringDocUtils.java
+++ b/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/utils/SpringDocUtils.java
@@ -565,13 +565,10 @@ public class SpringDocUtils {
 	 * @return the spring doc utils
 	 */
 	public SpringDocUtils initExtraSchemas() {
-		customClasses().put("java.time.Duration", PrimitiveType.STRING);
-		customClasses().put("java.time.LocalTime", PrimitiveType.STRING);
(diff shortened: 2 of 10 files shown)


bnasslahsen:
Merged, thanks @Mattias-Sehlstedt. Full suite is green on both lines.

Backported to `spring-boot-3` as well (2e7d271de), so 2.9.1 ships swagger-core 2.2.54 too — the `initExtraSchemas` entries and the `app115` / `app243` fixtures were identical on that line, so the cherry-pick applied cleanly.

Recorded in the changelog under **Changed**, flagging the visible consequence for users: `Duration` and `LocalTime` now carry `format: duration` / `format: partial-time` instead of a bare `string`.

didjoman:
Hi, 
I think this feature is creating a small regression when you are using SpringDoc 3.1.0 to generate OAS in version 3.0 instead of 3.1 (i.e. by setting the application property ``springdoc.api-docs.version=openapi_3_0``):

With SpringDoc 3.1.0 generating OAS in version 3.0, it gives this OAS: 
```json
"localTime": {
    "type": "string"
},
```

With SpringDoc 3.1.1 generating OAS in version 3.0, it gives this OAS:
```json
"localTime": {
    "properties": {
        "hour": {
            "format": "int32",
            "type": "integer"
        },
        "minute": {
            "format": "int32",
            "type": "integer"
        },
        "nano": {
            "format": "int32",
            "type": "integer"
        },
        "second": {
            "format": "int32",
            "type": "integer"
        }
    },
    "type": "object"
}
```

I think we should keep the extra schemas when ``springDocConfigProperties.getSpecVersion() == SpecVersion.V30``or something like that.

Best regards,
Alexandre

Mattias-Sehlstedt:
Thanks for raising the issue, I will investigate and see how to best restore the correct behavior of not exploding `LocalTime` (hopefully just some config that is need to ensure that swagger-core resolves it the same for OAS 3.0 and 3.1).

Mattias-Sehlstedt:
Hi @didjoman, I am unable to reproduce your issue, and to me they both work the same for OAS 3.1 and 3.0. Given that you get 
```json
"localTime": {
    "type": "string"
},
```
and not
```json
"localTime": {
    "type": "string"
    "format": "time-local"
},
```
then I believe that you might be using an older version of swagger-core. This would then result in your springdoc dropping the custom classes, while you do not pull in the new resolving from swagger-core since you are on an older version.

didjoman:
Hi, I am really sorry, indeed I had a fixed (older) version of swagger-core overriding springdoc one !
Thanks a lot

PR #3353, opened 2026-09-11
Mark Spring Data Page properties as required

dpkass:
Marks the standard properties of directly serialized Spring Data `Page` schemas as required. Spring Data always emits these fields, but the generated schema currently leaves them optional, causing client generators to model fields such as `content` and `totalPages` as optional.

This only applies to direct `Page` serialization. `VIA_DTO` serialization remains unchanged. Regression fixtures cover OpenAPI 3.0 and 3.1.


dpkass:
Closing this in favor of https://github.com/springdoc/springdoc-openapi/pull/3360. Marking every directly serialized `Page` property as required is not correct for `Page.unpaged()`, where that contract does not hold. The newer PR targets Spring Data’s stable `PagedModel` representation instead.

PR #3365, opened 2026-09-18
fix: make CollectionModelContentConverter conditional on LinkRelationProvider bean

### LINKED ISSUE #3364: SpringDocHateoasConfiguration breaks when there is no LinkRelationProvider bean
**Describe the bug**

`SpringDocHateoasConfiguration` assumes a `LinkRelationProvider` bean can be injected, but doesn't verify that such a bean actually exists.

**To Reproduce**

Spring Boot 4.1.1.

springdoc-openapi-starter-webmvc-api 3.1.1.

Not applicable.

`org.springdoc.core.configuration.SpringDocHateoasConfiguration#collectionModelContentConverter(HateoasHalProvider, LinkRelationProvider)` requires a `LinkRelationProvider` bean. This auto-configuration class verifies that the `LinkRelationProvider` class is present using `@ConditionalOnClass` (on multiple levels even, which seems redundant), but it never verifies that a bean of this type is present (using `@ConditionalOnBean`), so the auto-configuration fails when there is no `LinkRelationProvider` bean:

```
Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'modelConverterRegistrar' defined in class path resource [org/springdoc/core/configuration/SpringDocConfiguration.class]: Unsatisfied dependency expressed through method 'modelConverterRegistrar' parameter 0: Error creating bean with name 'collectionModelContentConverter' defined in class path resource [org/springdoc/core/configuration/SpringDocHateoasConfiguration.class]: Unsatisfied dependency expressed through method 'collectionModelContentConverter' parameter 1: No qualifying bean of type 'org.springframework.hateoas.server.LinkRelationProvider' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}
```

**Expected behavior**

- A clear and concise description of what you expected to happen.

I expect `SpringDocHateoasConfiguration` to not fail when the `LinkRelationProvider` class is present, but  there is no bean of this type.

- What is the expected result using OpenAPI Description (yml or json)?

Not applicable.

anupamchaubey:
## Description
`SpringDocHateoasConfiguration` currently uses `@ConditionalOnClass(LinkRelationProvider.class)`, but the `collectionModelContentConverter` bean method requires a `LinkRelationProvider` *bean*.

When the class is present on the classpath but no bean of that type exists, the application fails to start with:

## Changes
- Added `@ConditionalOnBean(LinkRelationProvider.class)` to the `collectionModelContentConverter` method.

This ensures the bean is only created when a `LinkRelationProvider` is actually available.

## Related Issue
Fixes #3364


Mattias-Sehlstedt:
Could we consider adding/extending the unit-test suite to also cover this case of application context configuration, so that we prevent regressions while also better showing the entire expected "application context suite" that might be used for the application. See for example the test suite that I modified in https://github.com/springdoc/springdoc-openapi/pull/3329.

anupamchaubey:
Thanks for the feedback. I've added a nested configuration guarded by @ConditionalOnBean(LinkRelationProvider.class) and extended the unit tests to cover both cases (bean present / bean absent). All tests in SpringDocHateoasConfigurationTest are now green.

